PR #2, opened 2020-11-11
Added pending jobs handling API to Runtime

katyo:
I added missing APIs to execute pending jobs in queue.
This is required for handling async stuff like promises and generators.

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -10,6 +10,8 @@ use std::{
     rc::Rc,
 };
 
+use crate::{context::Ctx, value};
+
 /// Opaque book keeping data for rust.
 pub struct Opaque {
     /// The registery, used to keep track of which registery values belong to this runtime.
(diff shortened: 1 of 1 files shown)


src/runtime.rs
```diff
@@ -133,6 +135,37 @@ impl Runtime {
...
+        let guard = self.inner.lock();
+        let res = unsafe { qjs::JS_IsJobPending(guard.rt) };
+        mem::drop(guard);
+        res != 0
+    }
+
+    /// Execute first pending job
+    ///
+    /// Returns context for executed job or none when queue is empty or error when exception thrown under execution.
+    pub fn execute_pending_job(&self) -> Result<Option<Ctx<'_>>, Error> {
```
DelSkayn, line 152: `Ctx` should not be available outside a callback handed to `Context::with` otherwise the runtime can be accessed outside the mutex gaurd, which is unsafe

DelSkayn: This functions should probably be a member of `Ctx`.

DelSkayn (changes requested):
These changes allows one to trigger undefined behavior from safe code. 
See comments for specifics.

src/runtime.rs
katyo: By design QuickJS has single queue per runtime, so this function cannot be member of `Ctx`.

katyo: My next pr #6 supersede this, I already changed returning value.

DelSkayn:
Closed, resolved in later pull request

PR #7, opened 2020-11-13
[WIP] Value conversion

katyo:
- [x] Missing `Array` set method
- [x] `Iterator` trait impl
  - [x] for `Array`
  - [x] for `Object` keys
  - [x] for `Object` props (kv-pairs)
- [x] `FromIteratorJs` trait to work with `Ctx`
- [x] `IteratorJs` trait which extends `Iterator` trait to support `Ctx`
  - [x] `IteratorJs::collect_js(ctx)` method to create `Array` and `Object` from iterator easy
- [x] `FromIteratorJs` impl
  - [x] for `Array`
  - [x] for `Object`
- [x] `ToJs` trait impls
  - [x] for primitive types
  - [x] for `Option`
  - [x] for `Result`
  - [x] for list-like collections
    - [x] for `Vec`
      The implementation conflicts with `ToJsMulti`.
      - [x] Add `MultiValueJs` as newtype for `Vec<Value>`
    - [x] for `VecDeque`
    - [x] for `LinkedList`
  - [x] for map-like collections
    - [x] for `HashMap`
    - [x] for `HashSet`
    - [x] for `BTreeMap`
    - [x] for `BTreeSet`
- [x] `FromJs` trait impls
  - [x] for primitive types
  - [x] for `Option`
  - [x] for `Result`
  - [x] for list-like collections
    - [x] for `Vec`
    - [x] for `VecDeque`
    - [x] for `LinkedList`
  - [x] for map-like collections
    - [x] for `HashMap`
    - [x] for `HashSet`
    - [x] for `BTreeMap`
    - [x] for `BTreeSet`
- [x] Support for variable number of multiple values using `RestValues` wrapper
  - [x] `FromJsMulti` trait impls
    - [x] for `RestValues<T>`
    - [x] for `(A, B, C, ..., RestValues<T>)`
  - [x] `ToJsMulti` trait impls
    - [x] for `RestValues<T>`
    - [x] for `(A, B, C, ..., RestValues<T>)`

### DIFF
--- a/src/value/array.rs
+++ b/src/value/array.rs
@@ -1,6 +1,6 @@
 use crate::{
     value::{self, rf::JsObjectRef},
-    Ctx, FromJs, Object, Result, ToJs, Value,
+    Ctx, FromIteratorJs, FromJs, Object, Result, ToJs, Value,
 };
 use rquickjs_sys as qjs;
 use std::{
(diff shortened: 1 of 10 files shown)


katyo:
Possibly we need to add some newtype for `Vec` to work around problem with `ToJsMulti`.

katyo:
I hope we can avoid using tuples for arguments passing at all.
I will try to apply technique used by [warp](https://github.com/seanmonstar/warp) crate.

katyo:
@DelSkayn 
I think we should rename traits `FromJsMulti` and `ToJsMulti` to `FromJsArgs` and `ToJsArgs` respectively to be more clear.

DelSkayn (approved):

katyo:
I'm in process of implementing procedural macros which gets easy bind Rust functions/modules/objects.

PR #20, opened 2020-11-24
Several meaningful improvements

katyo:
- [x] Support for user-defined memory allocators via `Allocator` trait (only available if `allocator` feature is enabled)
  - [x] Add `RustAllocator` implementation
  - [x] Use Rust global allocator by default if `rust-alloc` feature is enabled
- [x] Support for user-define module loaders (only available if `loader` feature is enabled)
  - [x] Module resolving via `Resolver` trait
    - [x] Add `FileResolver` which resolves the modules located on filesystem
    - [x] Add `BuiltinResolver` which resolves the modules built in program
  - [x] Module loading via `Loader` trait
    - [x] Add `ScriptLoader` which loads the JavaScript modules from files
    - [x] Add `NativeLoader` which loads the native compiled modules from shared objects or dynamic libraries
    - [x] Add `BuiltinLoader` which loads the JavaScript modules built in program
    - [x] Add `ModuleLoader` which loads the native compiled modules built in program
- [x] Support for native modules
  - [x] Added `ModuleDef` trait which allows initialize `Module` object when module loading
  - [x] Added state parameter to `Module` object which can be either `BeforeInit` or `AfterInit` (by default)
  - [x] Added `Module<BeforeInit>::add` method to add exported names before module instantiating
  - [x] Added `Module<AfterInit>::set` method to set actual exported values after module instantiating
  - [x] Added `Module::init::<impl ModuleDef>` helper function to get module initialization easy
  - [x] Added helper macro `module_init` to emit `js_init_module` function for specified `ModuleDef`
- [x] Module object impovements
  - [x] Removed inflexible `ExportList` iterator and `Module.export_list` method
  - [x] Added `ExportNamesIter<N>` which iterates over exported names and `Module::names` method
  - [x] Added `ExportEntriesIter<N, T>` which iterates over exported entries and `Module::entries` method
  - [x] Added method `Module::get` to get exported values by names directly
  - [x] Added method `Module::meta` to support `import.meta` object
  - [x] Added output type parameter to `Module::name`
- [x] Module loading examples
  - [x] Script module
  - [x] Native module
  - [x] Bundled script module
  - [x] Bundled native module
- [x] `IntoJs` trait impls for reference types
- [x] `Error` type without using `quick_error` macro
- [x] Avoid storing function class id in `Opaque` in `Runtime`
  This breaks ability for dynamic modules to be loaded by third-party runtimes which isn't based on `rquickjs`.
  Because class ids is global for entire program the function class id goes to be static.
- [x] Removed `FromJsArgs` trait in favor of `AsFunction` and `AsFunctionMut` traits which passes arguments to functions directly. This makes easy convert Rust functions to JS ones.
- [x] Removed `IntoJsArgs` trait in favor of `AsArguments` which does same in a little different more optimal way.
- [x] Removed `StaticFn` trait in favor of `AsFunction`
- [x] Added the wrappers for arguments:
  - [x] `This<T>` to get and pass `this` as an argument
  - [x] `Args<T>` to get and pass variable number of arguments
  - [x] `JsFn<F>` to pass rust function to JS via arguments or returns
- [x] Added the wrappers for functions:
  - [x] The `Method<F>` to convert methods to functions
- [x] Added features such as `max-args-N` where N in range `7..=16`
  It determines the maximum number of positional arguments
  which can be passed to and got by the functions.
  Compilation runs faster for smaller N and slower for bigger.
  By default `max-args-6` is used.
- [x] Added `FromJs` and `IntoJs` impls for tuples
  Tuples corresponds to an Arrays in JS.
- [x] Added `Persistent<T>` wrapper as a type-safe alternative for `RegisteryKey` which does not use `Runtime`'s `Opaque`
- [x] Support for user-defined classes
  - [x] Add `ClassDef` trait to define class for use in JS
  - [x] Add `class_def` macro to simplify class definition
  - [x] Add `Class<T>::register` method to register classes
  - [x] Add `IntoJs<T>` and `FromJs<&T>`, `FromJs<&mut T>` impls for everything which implements `ClassDef`
  - [x] Add `Constructor` wrapper which helps create constructors for classes
        The fn `Class::<T>::constructor(func)` can be used to wrap constructors easy.
  - [x] Add `Class<T>` as wrapper for `Object` with specific class `T`
  - [x] Add support for internal JS value references and gc_mark
- [x] Support for user-defined properties
  - [x] Add `Object::prop` method
  - [x] Add `AsProperty` trait
  - [x] Implement `AsProperty` for tuples to be able to define the following:
    - [x] Props without value via `()`
    - [x] Props with constant value via `(value,)`
    - [x] Props using getter only via `(getter,)`
    - [x] Props using getter and setter via `(getter, setter)`
  - [x] Add `AsPropertyFlag` trait
  - [x] Implement `AsPropertyFlag` for the following property flags:
    - [x] `NoConfigurable` (prop is configurable by default)
    - [x] `NoEnumerable` (prop is enumerable by default)
    - [x] `NoWritable` (props is writable by default)
    - [x] `(flag1, flag2, ..)` for combining multiple flags
  - [x] Implement `AsProperty` to add flags when defining properties:
    - [x] `(value, flags)`
    - [x] `(getter, flags)`
    - [x] `(getter, setter, flags)`


katyo:
I think the `exports` feature is not so necessary.
Related functionality just could be enabled implicitly, there is no any reasons to disable it.

DelSkayn:
> I think the `exports` feature is not so necessary.
> Related functionality just could be enabled implicitly, there is no any reasons to disable it.

The exports feature is there mostly so that users can have an unmodified quickjs if they wanted to (and because I am not fully sure that the alterations made are fully sound). So I think there nice to have.

I am currently pretty busy btw so I don't have much time to work on this library.
Let me know if you need anything from me.

PR #21, opened 2021-02-01
Async executor related fixes

katyo:
This is a work in progress PR.
I tries to fix bugs related to async execution which still appears especially when running in async-std runtime.
That PR should be merged before releasing v0.1.0


katyo:
Something strange happens on i686.
The `delayed_fn_swarm` test runs infinitely on debug build.
When I run it locally with `--release` it just works as expected.

PR #28, opened 2021-08-22
Re-organizing crates and update dependencies

katyo:


chitoyuu:
@katyo I've tried this and it's working great! :tada: It seems that the `hand_written` versions of the native modules are still using `JsFn`, though. Building `module-loader` with `--no-default-features` resulted in the following error:

```
error[E0432]: unresolved import `rquickjs::JsFn`
 --> examples/module-loader/src/bundle/hand_written.rs:1:30
  |
1 | use rquickjs::{Created, Ctx, JsFn, Loaded, Module, ModuleDef, Native, Result};
  |                              ^^^^ no `JsFn` in the root

For more information about this error, try `rustc --explain E0432`.
error: could not compile `module-loader` due to previous error
```

It seems that you still need to update the identifiers in those examples, and perhaps add `--no-default-features` builds to the CI so this could be catched automatically in the future.

core/src/context.rs
```diff
@@ -91,7 +91,7 @@ impl Context {
 
     #[cfg(feature = "parallel")]
     fn reset_stack(&self) {
-        unsafe { qjs::JS_ResetStackPointer(self.ctx) };
+        unsafe { qjs::JS_UpdateStackTop(qjs::JS_GetRuntime(self.ctx)) };
```
DelSkayn, line 94: Somewhat pedantic note.
Retrieving the runtime pointer this way is unsafe since if the function is called before the runtime is locked it could result in a race condition.
Not that this function is used ever before acquiring a lock, but it might be somewhat more ideologically pure to have this function take the locked runtime as an argument.

katyo: Ohh, seems I planned move thi function to `Inner` of `Runtime` but forgot to do it :)

PR #29, opened 2021-09-24
Update quickjs version

DelSkayn:
This branch includes changes to update rquickjs to a newer version of quickjs.

Also includes some minor fixes to deal with warnings generated in newer versions of the rust compiler.

PR #30, opened 2021-10-07
Make examples work again

chitoyuu:
Hello!

Firstly I must say that at a quick glance, this seems to be a much more comprehensive and promising effort than the other QuickJS bindings available on crates. Kudos to you!

Right now, it seems that the examples are broken on `master`. I have managed to make them work again to my best understanding, but I'm not sure if the way I did it in `examples/module-loader/src/main.rs` is the best. Some help would be appreciated.

Thanks!


chitoyuu:
@katyo Perfect! I think this can be closed then. May I ask what's blocking #28 and if I could do anything to help?

katyo:
@chitoyuu Just now I integrated your fixes and I in process of integrating #29.
So when it will be ready you may help with testing PR #28 before merging.

PR #49, opened 2022-01-26
Added initial support for ArrayBuffer and TypedArray(s)

katyo:
- [x] Added `ArrayBuffer` interface type
- [x] Added `TypedArray<T>` interface type
- [x] Implemented type conversions
- [x] Updated docs


paulrouget:
This appears to work well! Not sure about what tests to add, but code looks good.

PR #50, opened 2022-02-23
[WIP] Add Context::lock

arthurprs:
This change adds `Context::lock`, which can be useful in some scenarios.

Eg. an async function like the one below can't use the `::with` function.

```rust
    pub async fn fold<'a>(
        &self,
        fn_name: &str,
        global_context: &serde_json::Value,
        stream: impl Stream<Item = Item>,
    ) -> anyhow::Result<serde_json::Value> {
        let guard = self.context.lock();
        let ctx = guard.get();

        let globals = ctx.globals();
        globals.set("context", JsonValueRef(global_context))?;

        let js_fn = globals.get::<_, rquickjs::Function>(fn_name)?;
        let mut user_arg = rquickjs::Value::new_undefined(ctx);
        futures::pin_mut!(stream);
        while let Some(item) = stream.next().await {
            user_arg = js_fn.call((JsonValueRef(&item),))?;
        }
        Ok(JsonValue::from_js(ctx, user_arg)?.0)
    }
```

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -1,5 +1,5 @@
 use crate::{qjs, Error, Result, Runtime};
-use std::mem;
+use std::{borrow::Cow, mem};
 
 mod builder;
 pub use builder::{intrinsic, ContextBuilder, Intrinsic};
@@ -137,8 +137,28 @@ impl Context {
         let guard = self.rt.inner.lock();
         guard.update_stack_top();
         CtxGuard {
-            context: self,
+            context: Cow::Borrowed(self),
             guard: mem::ManuallyDrop::new(guard),
+        }
+    }
+
+    /// Transforms a [`Context`] in a [`OwnedCtxGuard`] for manipulating and using javascript
+    /// objects and scripts, blocking the current thread until it is able to do so.
+    ///
+    /// This function will block the local thread until it is available to acquire
+    /// the lock. Upon returning, the thread is the only thread with the lock
+    /// held. An RAII guard is returned to allow scoped unlock of the lock. When
+    /// the guard goes out of scope, the runtime will be unlocked.
+    pub fn owned_lock(self) -> OwnedCtxGuard {
+        let guard = self.rt.inner.lock();
+        guard.update_stack_top();
+        CtxGuard {
+            // Safety: here we force the guard to have the static lifetime by transmuting.
+            // This is safe because Context and it's contents refer to a stable addresses, even
+            // when moved. Furthermore, we are guaranteeing that those addresses contents stay
+            // valid by moving self into the CtxGuard as well.
+            guard: mem::ManuallyDrop::new(unsafe { mem::transmute(guard) }),
+            context: Cow::Owned(self),
         }
     }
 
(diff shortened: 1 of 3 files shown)


katyo:
@arthurprs Seems you forgot export `CtxGuard` to get it visible outside of this crate.

arthurprs:
@katyo Indeed! If you don't mind, maybe hold off merging it for a couple days. I'm still trying to integrate this crate in my project (thank you :bow:) and I'm still hitting some problems. I might need to add an "OwnedGuard" kind of thing.

katyo:
@arthurprs,
Course, could you mark this PR as [WIP] or Draft:?
You can remove this mark and request review from me when everything is ready.

DelSkayn:
This pull request as it is currently is unfortunately unsound.
As example see the following program, which compiles at the latest commit in this pr:
```rust
use rquickjs::{Context, Runtime, Value};

fn main() {
    let rt_1 = Runtime::new().unwrap();
    let rt_2 = Runtime::new().unwrap();
    let ctx_1 = Context::full(&rt_1).unwrap();
    let ctx_2 = Context::full(&rt_2).unwrap();
    let guard_1 = ctx_1.lock();
    let guard_2 = ctx_2.lock();
    let val: Value = guard_1.get().globals().into();
    // Unsound!!
    guard_2.get().globals().set("test", val).unwrap();
}
```
The library was designed with the possibility of multiple runtimes in mind.
As exchanging objects between two different runtimes can lead to undefined behaviour, we need to ensure that quickjs values belong to the right runtime before they are used with a certain `Context` object.

In order to prevent testing for the right runtime each time an object is passed into the library, I copied the with scope design from the library rlua, which I belief uses this design for the same reason.
The following code, adjusted from one of the compile tests,  will not compile:
```rust
use rquickjs::{Context, Runtime, Value};

fn main() {
    let rt_1 = Runtime::new().unwrap();
    let rt_2 = Runtime::new().unwrap();
    let ctx_1 = Context::full(&rt_1).unwrap();
    let ctx_2 = Context::full(&rt_2).unwrap();
    ctx_1.with(|ctx_1| {
        // Error here: "cannot infer an appropriate lifetime for lifetime parameter `'js`"
        let val: Value = ctx_1.globals().into();
        ctx_2.with(|ctx_2| {
            ctx_2.globals().set("foo", val);
        })
    })
}
```
By using this method the lifetimes of the objects inside the context can both, not escape the context, and can also not be used with a scope with a smaller lifetime, essentially binding the objects to that specific context.
The makes it possible to safely use objects without checking if they belong the right runtime within the `with` scope.

I thought this was not possible with a lock design, although I am not 100% certain on that.

If we would want to implement drop guards for obtaining `Ctx` objects we would also need to implement runtime checks everywhere where an object is passed into the library. 
Which I think would both increase complexity and might decrease performance. 
So if it is possible I think a other solution then a lock, if we could find one, might be better.

DelSkayn:
PS: Altough I have played this pr's  and have not been able to make it work, I might absolutely be wrong that there is no way to do the compile time runtime check with locks. 
So if I am wrong please correct me.

arthurprs:
I agree with your assessment. I think of this changeset only the last commit make sense. I'll go ahead and close this and make a follow up PR with only that.

PR #51, opened 2022-03-14
Enable usage of `Value` in `Persistent` and guard against restoring in the wrong `Runtime`

arthurprs:
Take `Into<Value>` instead of `AsRef<Value>` in `Persistent::save` so it's usable with `Value`s

This is mainly due to Rust limitation that `AsRef<T>` is not implemented for neither `T` nor `&T`.

Note that this could be a breaking change, but otherwise `Persistent::save` is not usable with `Value<'js>`.

Guard against restoring `Persistent` in a different `Runtime`. Also adds a `#[non_exhaustive]` tag to the `Error` enum so adding variants isn't a breaking change in the future.

This is a follow up of #50. #50 was ultimately the wrong approach, that I only really took because I was unable to make Persistent<Value> work well, but this change should improve things.

### DIFF
--- a/core/src/persistent.rs
+++ b/core/src/persistent.rs
@@ -108,9 +108,10 @@ impl<T> Persistent<T> {
     /// Save the value of an arbitrary type
     pub fn save<'js>(ctx: Ctx<'js>, val: T) -> Persistent<T::Target>
     where
-        T: AsRef<Value<'js>> + Outlive<'static>,
+        T: Into<Value<'js>> + Outlive<'static>,
     {
-        let value = val.as_ref().value;
+        let val = val.into();
+        let value = val.value;
         mem::forget(val);
         let rt = unsafe { qjs::JS_GetRuntime(ctx.ctx) };
 
@@ -210,5 +211,23 @@ mod test {
             func.call((0,)).unwrap()
         });
         assert_eq!(res, 1);
+    }
+
+    #[test]
+    fn persistent_value() {
+        let rt = Runtime::new().unwrap();
+        let ctx = Context::full(&rt).unwrap();
+
+        let persistent_v = ctx.with(|ctx| {
+            let v: Value = ctx.eval("1").unwrap();
+            Persistent::save(ctx, v)
+        });
+
+        ctx.with(|ctx| {
+            let v = persistent_v.clone().restore(ctx).unwrap();
+            ctx.globals().set("v", v).unwrap();
+            let eq: Value = ctx.eval("v == 1").unwrap();
+            assert!(eq.as_bool().unwrap());
+        });
     }
 }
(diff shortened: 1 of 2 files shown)


core/src/persistent.rs
```diff
@@ -188,6 +199,26 @@ impl<T> Eq for Persistent<T> {}
...
+            Persistent::save(ctx, v)
+        });
+
+        let rt2 = Runtime::new().unwrap();
+        let ctx = Context::full(&rt2).unwrap();
+        ctx.with(|ctx| {
+            assert!(matches!(
+                persistent_v.clone().restore(ctx),
+                Err(Error::DifferentRuntime)
+            ));
```
katyo, line 218: Please use `#[should_panic]` instead:
```rust
#[should_panic(expected = "Expected error reason")]
fn test() {
...
```

core/src/result.rs
```diff
@@ -65,6 +66,8 @@ pub enum Error {
         name: StdString,
         message: Option<StdString>,
     },
+    /// Error when restoring a Persistent in a runtime other than the original runtime.
+    DifferentRuntime,
```
katyo, line 70: I think the more proper variant should sounds like `IrrelevantRuntime`.

core/src/result.rs
```diff
@@ -355,6 +358,7 @@ impl Display for Error {
                 "IO Error: ".fmt(f)?;
                 error.fmt(f)?;
             }
+            DifferentRuntime => "Restoring Persistent in a different runtime".fmt(f)?,
         }
```
katyo, line 362: `"Restoring persistent value in a not relevant runtime"`

core/src/result.rs
```diff
@@ -355,6 +358,7 @@ impl Display for Error {
                 "IO Error: ".fmt(f)?;
                 error.fmt(f)?;
             }
+            DifferentRuntime => "Restoring Persistent in a different runtime".fmt(f)?,
```
katyo, line 361: "Restoring persistent value in a not relevant runtime"

core/src/persistent.rs
arthurprs: Done

core/src/result.rs
arthurprs: `UnrelatedRuntime` sounds good to me. Pushed changes for that.

katyo:
Hmm, seems CI hasn't runs on this RP. Also I cannot run it manually.

katyo:
@arthurprs Seems I forget add `impl AsRef<Value> for Value`. So we don't need to make breaking changes.

katyo:
@arthurprs Could you fix and rebase this PR?

katyo:
Superseded by #54

PR #52, opened 2022-03-15
Add interrupt handler api

arthurprs:
It's an useful API for interrupting embedded code.

### DIFF
--- a/core/src/runtime.rs
+++ b/core/src/runtime.rs
@@ -1,5 +1,5 @@
 use crate::{get_exception, qjs, Ctx, Error, Function, Mut, Ref, Result, Weak};
-use std::{any::Any, ffi::CString, mem};
+use std::{any::Any, ffi::CString, mem, panic};
 
 #[cfg(feature = "futures")]
 mod async_runtime;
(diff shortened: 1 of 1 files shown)


core/src/runtime.rs
```diff
@@ -214,6 +217,39 @@ impl Runtime {
...
+        }
+
+        let mut guard = self.inner.lock();
+        unsafe {
+            qjs::JS_SetInterruptHandler(
+                guard.rt,
+                handler.as_ref().map(|_| interrupt_handler_trampoline as _),
+                qjs::JS_GetRuntimeOpaque(guard.rt),
+            );
+            guard.get_opaque_mut().interrupt_handler = handler;
```
katyo, line 249: Could this line be moved out of unsafe block?

core/src/runtime.rs
```diff
@@ -214,6 +217,39 @@ impl Runtime {
         WeakRuntime(Ref::downgrade(&self.inner))
     }
 
+    pub fn set_interrupt_handler(&self, handler: Option<Box<dyn FnMut() -> bool + 'static>>) {
```
katyo, line 220: Please describe this function using doc-comment.

arthurprs: Ops, I overlooked it. It's included now.

PR #54, opened 2022-03-19
Enable usage of `Value` in `Persistent` and guard against restoring in the unrelated `Runtime`

katyo:
Supersedes #51


arthurprs:
Could we cut out a release that includes the latest PRs?

katyo:
@arthurprs Of course.
But I would like fix windows-gnu targets support on CI first.

PR #65, opened 2022-05-30
Add `Ctx::eval_with_options` and `Ctx::eval_file_with_options`.

### LINKED ISSUE #64: Strict mode is always enforced / how to run non "use strict"; code?
## Problem

I'm trying to execute the following code snippet:

```rust

    ctx.with(|ctx| {
        let out: String = ctx
            .eval(
                r#"
                function test() {
                    foo = "bar";
                    return foo;
                }

                test()
            "#,
            )
            .unwrap();
        println!("{}", out);
    });
```

```text
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Exception { message: "'foo' is not defined", file: "", line: -1, stack: "    at test (eval_script:3)\n    at <eval> (eval_script:7)\n" }'
```

While I'm aware that it would be better to write `var foo = "bar"`, I'm dealing with code that is is provided by other users. Using Node.js, the above works fine as long `"use strict";` is not used.

How can I make the above code work using the `rquickjs` library?

## Observations

The `eval` method sets the `qjs::JS_EVAL_FLAG_STRICT` flag, as does the `compile` (indirectly, as it calls `Module::new` which sets the same flag).

* https://github.com/DelSkayn/rquickjs/blob/master/core/src/context/ctx.rs#L61=
* https://github.com/DelSkayn/rquickjs/blob/master/core/src/context/ctx.rs#L91=
* https://github.com/DelSkayn/rquickjs/blob/master/core/src/value/module.rs#L188=

Reading https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode, strict mode is supposed to be opt-in. Only for modules and classes strict mode is always enforced.

> ECMAScript 2015 introduced [JavaScript modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) and therefore a 3rd way to enter strict mode. The entire contents of JavaScript modules are automatically in strict mode, with no statement needed to initiate it.

> All parts of ECMAScript [classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes) are strict mode code, including both [class declarations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#class_declarations) and [class expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#class_expressions) — and so also including all parts of class bodies.

## Possible solution?

Should the `eval` method set the `qjs::JS_EVAL_FLAG_STRICT` flag? Thus instead of:

```rust
    /// Evaluate a script in global context
    pub fn eval<V: FromJs<'js>, S: Into<Vec<u8>>>(self, source: S) -> Result<V> {
        let file_name = unsafe { CStr::from_bytes_with_nul_unchecked(b"eval_script\0") };
        let flag = qjs::JS_EVAL_TYPE_GLOBAL | qjs::JS_EVAL_FLAG_STRICT;
        V::from_js(self, unsafe {
            let val = self.eval_raw(source, file_name, flag as i32)?;
            Value::from_js_value(self, val)
        })
    }
```

Shouldn't it be:

```rust
    /// Evaluate a script in global context
    pub fn eval<V: FromJs<'js>, S: Into<Vec<u8>>>(self, source: S) -> Result<V> {
        let file_name = unsafe { CStr::from_bytes_with_nul_unchecked(b"eval_script\0") };
        let flag = qjs::JS_EVAL_TYPE_GLOBAL;
        V::from_js(self, unsafe {
            let val = self.eval_raw(source, file_name, flag as i32)?;
            Value::from_js_value(self, val)
        })
    }
```

I assume that in this case strict mode can still be enforced by the user by using `"use strict";`.

brocaar:
Closes #64.

### DIFF
--- a/core/src/context/ctx.rs
+++ b/core/src/context/ctx.rs
@@ -26,6 +26,28 @@ pub struct EvalOptions {
     pub global: bool,
     /// Force 'strict' mode.
     pub strict: bool,
+    /// Don't include the stack frames before this eval in the Error() backtraces.
+    pub backtrace_barrier: bool,
+}
+
+impl EvalOptions {
+    fn to_flag(&self) -> i32 {
+        let mut flag = if self.global {
+            qjs::JS_EVAL_TYPE_GLOBAL
+        } else {
+            qjs::JS_EVAL_TYPE_MODULE
+        };
+
+        if self.strict {
+            flag |= qjs::JS_EVAL_FLAG_STRICT;
+        }
+
+        if self.backtrace_barrier {
+            flag |= qjs::JS_EVAL_FLAG_BACKTRACE_BARRIER;
+        }
+
+        flag as i32
+    }
 }
 
 impl Default for EvalOptions {
@@ -33,6 +55,7 @@ impl Default for EvalOptions {
         EvalOptions {
             global: true,
             strict: true,
+            backtrace_barrier: false,
         }
     }
 }
@@ -85,18 +108,8 @@ impl<'js> Ctx<'js> {
     ) -> Result<V> {
         let file_name = unsafe { CStr::from_bytes_with_nul_unchecked(b"eval_script\0") };
 
-        let mut flag = if options.global {
-            qjs::JS_EVAL_TYPE_GLOBAL
-        } else {
-            qjs::JS_EVAL_TYPE_MODULE
-        };
-
-        if options.strict {
-            flag |= qjs::JS_EVAL_FLAG_STRICT;
-        }
-
         V::from_js(self, unsafe {
-            let val = self.eval_raw(source, file_name, flag as i32)?;
+            let val = self.eval_raw(source, file_name, options.to_flag())?;
             Value::from_js_value(self, val)
         })
     }
@@ -120,18 +133,8 @@ impl<'js> Ctx<'js> {
                 .into_owned(),
         )?;
 
-        let mut flag = if options.global {
-            qjs::JS_EVAL_TYPE_GLOBAL
-        } else {
-            qjs::JS_EVAL_TYPE_MODULE
-        };
-
-        if options.strict {
-            flag |= qjs::JS_EVAL_FLAG_STRICT;
-        }
-
         V::from_js(self, unsafe {
-            let val = self.eval_raw(buffer, file_name.as_c_str(), flag as i32)?;
+            let val = self.eval_raw(buffer, file_name.as_c_str(), options.to_flag())?;
             Value::from_js_value(self, val)
         })
     }
--- a/core/src/context/ctx.rs
+++ b/core/src/context/ctx.rs
@@ -58,7 +58,7 @@ impl<'js> Ctx<'js> {
     /// Evaluate a script in global context
     pub fn eval<V: FromJs<'js>, S: Into<Vec<u8>>>(self, source: S) -> Result<V> {
         let file_name = unsafe { CStr::from_bytes_with_nul_unchecked(b"eval_script\0") };
-        let flag = qjs::JS_EVAL_TYPE_GLOBAL | qjs::JS_EVAL_FLAG_STRICT;
+        let flag = qjs::JS_EVAL_TYPE_GLOBAL;
         V::from_js(self, unsafe {
             let val = self.eval_raw(source, file_name, flag as i32)?;
             Value::from_js_value(self, val)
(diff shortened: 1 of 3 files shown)


katyo:
@brocaar 
Such changes breaks back compatibility.
Maybe a better solution is additional non-strict `eval` methods (i.e. `eval_nostrict`).

brocaar:
@katyo I'll update this PR with `eval_nostrict`

brocaar:
@katyo please see the above commit.

brocaar:
> Maybe a more proper solution is introducing EvalOptions type and functions like eval_with_options.

Do we know which other options we want to provide (in the future)? I did find these `JS_EVAL_...` flags:

```
pub const JS_EVAL_TYPE_GLOBAL: u32 = 0;
pub const JS_EVAL_TYPE_MODULE: u32 = 1;
pub const JS_EVAL_TYPE_DIRECT: u32 = 2;
pub const JS_EVAL_TYPE_INDIRECT: u32 = 3;
pub const JS_EVAL_TYPE_MASK: u32 = 3;
pub const JS_EVAL_FLAG_STRICT: u32 = 8;
pub const JS_EVAL_FLAG_STRIP: u32 = 16;
pub const JS_EVAL_FLAG_COMPILE_ONLY: u32 = 32;
pub const JS_EVAL_FLAG_BACKTRACE_BARRIER: u32 = 64;
```

However, I'm not sure if these should all be exposed to end-users.

Looking at the quickjs source:

```
/* JS_Eval() flags */
#define JS_EVAL_TYPE_GLOBAL   (0 << 0) /* global code (default) */
#define JS_EVAL_TYPE_MODULE   (1 << 0) /* module code */
#define JS_EVAL_TYPE_DIRECT   (2 << 0) /* direct call (internal use) */
#define JS_EVAL_TYPE_INDIRECT (3 << 0) /* indirect call (internal use) */
#define JS_EVAL_TYPE_MASK     (3 << 0)

#define JS_EVAL_FLAG_STRICT   (1 << 3) /* force 'strict' mode */
#define JS_EVAL_FLAG_STRIP    (1 << 4) /* force 'strip' mode */
/* compile but do not run. The result is an object with a
   JS_TAG_FUNCTION_BYTECODE or JS_TAG_MODULE tag. It can be executed
   with JS_EvalFunction(). */
#define JS_EVAL_FLAG_COMPILE_ONLY (1 << 5)
/* don't include the stack frames before this eval in the Error() backtraces */
#define JS_EVAL_FLAG_BACKTRACE_BARRIER (1 << 6)
```

I think the `_DIRECT` and `_INDIRECT` flags should not be exposed. Please let me know what you think about the other flags.

> Of course, the name of function too long. But variants like eval_as sounds misleading in my opinion.

I agree that `eval_as` is misleading. It sounds more like a typing thing. To me `eval_with_options` is fine.

brocaar:
Looks like CI errors are unrelated?

> Can't use 'tar -xzf' extract archive file: /home/runner/work/_actions/_temp_aca0f43c-86e0-4c9a-a164-7c827038b2ce/cb8356a9-0b53-4f57-80c8-93f8cea8e4b1.tar.gz. return code: 2.

katyo:
@brocaar Hmm, CI looks broken. I can test it myself and merge.

core/src/context/ctx.rs
```diff
@@ -55,10 +72,28 @@ impl<'js> Ctx<'js> {
...
+        source: S,
+        options: EvalOptions,
+    ) -> Result<V> {
         let file_name = unsafe { CStr::from_bytes_with_nul_unchecked(b"eval_script\0") };
-        let flag = qjs::JS_EVAL_TYPE_GLOBAL | qjs::JS_EVAL_FLAG_STRICT;
+        let mut flag = 0;
+
+        if options.global {
+            flag |= qjs::JS_EVAL_TYPE_GLOBAL;
+        }
```
katyo, line 91: In that context the opposite of `global` flags is `module` so in my opinion this block should looks like the following:
```rust
flag |= if options.global { qjs::JS_EVAL_TYPE_GLOBAL } else { qjs::JS_EVAL_TYPE_MODULE };
```
Otherwise this block is useless because the `JS_EVAL_TYPE_GLOBAL` flag is zero.
Other variants is out of scope because it is for internal use only.

katyo:
I think `backtrace_barrier` flag may be useful for eval. I don't know which `strip` flag means.

brocaar:
@katyo and I've updated the PR again :-) Once this is merged, would you mind creating a new release of this library?

katyo:
@brocaar Of course (:

PR #90, opened 2022-09-26
Removing mutable access to ClassDef.

### LINKED ISSUE #83: `Class::try_mut` is unsound
The `Class::try_mut` method can be used to obtain aliasing mutable references to the same `ClassDef` object.

```rust
let cls: Class<_> = \* create here *\

// Create a reference copy to the same object
let obj = cls.to_object();
let obj_1 = obj.clone()

let cls = Class::from_object(obj);
let cls_1 = Class::from_object(obj_1);
// There are several other ways to obtain a copy

// Use try_mut to obtain a mutable reference.
let a = cls.try_mut().unwrap();
let b = cls_1.try_mut().unwrap();
// `a` and `b` are both a mutable reference to the same `ClassDef` object which is UB.
```
Class objects and their contained `ClassDef` object behave similarly as an `Rc` or and `Arc` which also do not allow obtaining a mutable reference to their contained object as the lifetime of the object contained is inherently shared. 

I see three possible solutions to this problem:

1. **Do not allow multiple class object pointing to the same object to exist.**
This seem extremely restrictive and rather hard to maintain. Obtaining multiple references to the same object in javascript is trivial.
As there is no good method to track class objects on the javascript side this would probably mean that `ClassDef` objects can only be access while on the rust side and once passed to the javascript side should no longer be accessible.

2. **Enforce proper access pattern to the `ClassDef` object.**
This solution has the `ClassDef` object wrapped in some `RefCell` like structure which regulates access pattern preventing multiple mutable access to the same object.
This solution minimizes the surface area of the breaking changes only `try_mut` and `try_ref` would return a guard object instead of a naked reference. The `bind` macros could probably remain the same. This would probably also change the behavior of some of the library as it is probably required that certain functions panic when access patterns are violated.

3. **Force all access to the `ClassDef` object to be a immutable reference.**
To me this seems the most clean solution on the library implementation side. Mutable access to a `ClassDef` object is then forced to be handled by the library user through objects like `Cell` and `Mutex`. This removes unnecessary overhead if mutable access is not required when compared to the previous solution.
This change to the library is, however, a breaking change with a pretty large surface area, as not only the `try_mut` method should be removed, but also mutable access in the `bind` macros as well as possibly other API which relies on mutable access to the `ClassDef` object.
We could prevent mostly breaking the `bind` macros by allowing the macro implementation to be aware of cell like objects to implement methods which mutate the `ClassDef` object.

DelSkayn:
PR to fix #83.

### DIFF
--- a/core/src/class.rs
+++ b/core/src/class.rs
@@ -61,12 +61,6 @@ pub use refs::{HasRefs, RefsMarker};
 ///     }
 /// }
 ///
-/// impl<'js> FromJs<'js> for &'js mut MyClass {
-///     fn from_js(ctx: Ctx<'js>, value: Value<'js>) -> Result<Self> {
-///         MyClass::from_js_mut(ctx, value)
-///     }
-/// }
-///
 /// impl<'js> FromJs<'js> for MyClass {
 ///     fn from_js(ctx: Ctx<'js>, value: Value<'js>) -> Result<Self> {
 ///         MyClass::from_js_obj(ctx, value)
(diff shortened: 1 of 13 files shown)


katyo:
@DelSkayn
I'm sorry for the long absence. Generally this changes looks good to me.
Are you tested mutex with classes (with parallel feature enabled)?

katyo:
@DelSkayn Looks like we ready to merge it and issue v0.2.

DelSkayn:
This can indeed be merged, The only thing is that this pr drastically worsens the ergonomics of the library. 
I had hoped to be able to look at implementing a method to automatically wrap ClassDef in a RefCell by opting in via some macro attribute. I just haven't had the time however so in the mean time it is probably fine to go for v0.2

PR #115, opened 2023-03-02
Some fixes having to do with classes.

### LINKED ISSUE #92: Registering class with prototype in multiple contexts of the same runtime not possible.
It appears that registering a bound struct in multiple contexts of the same runtime is not possible.

Reproduction: https://github.com/bspot/repro_register_class_twice

I'm not sure, but I believe the problem is that `Class::register` skips associating the prototype with class when the class itself is already registered with the runtime. However, the prototype needs to be associated with class for each context individually.

DelSkayn:
This changes `ClassId` to instead of requiring a static mutable for implementing `ClassDef` it now only needs an immutable static. `ClassId` can will now initialize itself only once when its the id is read first. This prevents multiple calls to register from altering the class id. 

This also fixes #92. Once a class was registered in a runtime its prototype was no longer being set. This PR fixes that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-docs
 target
 .ccls-cache
 Cargo.lock
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+docs
 target
 .ccls-cache
 Cargo.lock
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+docs
```
katyo, line 1: ???

katyo (approved):

.gitignore
DelSkayn: Oh, yeah, I'll remove that. I had some notes in my local repository.

Issue #59, opened 2022-05-02
Cannot use `bind` or `apply` or other `Function.prototype` methods on a native function


jeromegn:
For example:

```rust
fn print(msg: String) {
    println!("{}", msg);
}

fn main() {
    let mut rt = Runtime::new().unwrap();
    let mut ctx = Context::full(&rt).unwrap();

    ctx.with(|ctx| {
        ctx.globals().set("print", Func::new("print", print)).unwrap();
        ctx.eval::<(), _>(r#"print.bind(this, "hello")();"#).unwrap();
    })
}
```

Results in this error:

```
Exception { message: "not a function", file: "", line: -1, stack: "    at <eval> (eval_script:2)\n" }
```

DelSkayn:
This is somewhat by design. In quickjs C functions (or rust functions) have their prototype set to the object prototype. There is no real good way to set the prototype of function objects to the function prototype from which the `bind` and `apply` functions are derived. 

The best way to solve this issue is to wrap C/rust functions on the javascript side with a javascript function.

katyo:
@DelSkayn We can try to use Function.prototype as prototype for Rust functions but I'm not sure.

DelSkayn:
> @DelSkayn We can try to use Function.prototype as prototype for prototype of Rust functions but I'm not sure.

I have looked into setting the prototype of native functions to Function.prototype. The problem I encountered is that there is no good way to access the function prototype object from rust as quickjs does not allow you to access it. Which might mean that that prototype is only meant for javascript functions and not for C functions but I am not sure about that.

As I see it right now there are three options:
1. Patching quickjs so that we can access prototype object.
2. Doing something like `let func_proto = ctx.eval("Function.prototype")` at library initialization and than use the returned object as a prototype for functions.
3. When creating a native function wrapping it with `(...args) => nativeFunc(...args)` in javascript.

Option 1 and 2 felt too much like a hack to me and option 3 seems like it could cause some overhead as I did not like the requirement to have to run `Ctx.eval` for every function created. 
So it seemed to me that the best solution was to have users just wrap functions if they need defined prototype methods as with something like:
```rust
fn wrap_func<'js>(ctx: Ctx<'js>, func: Function<'js>) -> rquickjs::Result<Function<'js>> {
    ctx.eval::<Function, _>("(f) => (...args) => f(...args)")?
        .call((func,))?
}
```

PR #118, opened 2023-03-13
Switch rust defined function to use `JS_NewCFunctionData`

DelSkayn:
This PR switches rust function from being created as a class with a call field to instead use `JS_NewCFunctionData`. By creating functions this way the function will properly inherit the function prototype and have function prototype methods like `apply` and `bind` defined.

This might cause slightly more overhead than the previous way of creating function objects as this methods requires two object to be created: the function itself and a finalizer object which drops the function data.

This PR fixes issue #59.

### DIFF
--- a/core/src/value/function.rs
+++ b/core/src/value/function.rs
@@ -34,7 +34,6 @@ impl<'js> Function<'js> {
             Self::from_js_value(ctx, func)
         };
         F::post(ctx, &func)?;
-        func.set_length(F::num_args().start)?;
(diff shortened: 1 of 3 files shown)


katyo:
@DelSkayn 
Are you tried the solution with creating function via `JS_NewObjectProtoClass` with `Function.prototype` as a proto?
I don't sure but seems it may be simpler.

DelSkayn:
Closed in favor of #120

PR #119, opened 2023-03-14
Promise rejection handler

richarddd:
## Usage

```rust
runtime.set_promise_rejection_tracker(Some(Box::new(|value|{
//value is rejection from uncaught promise
})));
```

## Background

QuickJS has JS_SetHostPromiseRejectionTracker.

However, it will report falsely report promises that are later handled.
To work around this issue, the uncaught promises are tracked and stored in a HashSet. If its handled in a later .then or .catch its removed from the Set.

The callback is then executed before runtime is dropped. This is not ideal as we would only discover uncaught promises when we drop the runtime.

Is there any smarter mechanism to run the callback without reporting false rejections?

## Js to test
_Should report uncaught promises:_
```javascript
const main = async () => {
  throw new Error("error");
};

main(); 
```

_Should not report uncaught promises:_
```javascript
const main = async () => {
  throw new Error("error");
};

main().then(()=>{},()=>{})
```

### DIFF
--- a/core/src/runtime.rs
+++ b/core/src/runtime.rs
@@ -1,10 +1,11 @@
-use crate::{get_exception, qjs, Ctx, Error, Function, Mut, Ref, Result, Weak};
-use std::{any::Any, ffi::CString, mem, panic};
+use crate::{get_exception, qjs, Ctx, Error, Function, Mut, Ref, Result, Weak, Value};
+use std::{any::Any, ffi::CString, mem, panic, collections::HashSet, hash::{Hash, Hasher}};
 
 #[cfg(feature = "futures")]
 mod async_runtime;
 #[cfg(feature = "futures")]
 pub use async_runtime::*;
+use rquickjs_sys::{JSContext, JSValue};
 #[cfg(feature = "futures")]
 mod async_executor;
 #[cfg(feature = "futures")]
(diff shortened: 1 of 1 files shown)


richarddd:
@katyo @DelSkayn whats your take on this?

katyo:
@richarddd Could you fix formatting (with `cargo fmt`) and rebase this PR on the current master?

DelSkayn:
Looking into the source code of quickjs it seems that the behavior of reporting promises which are handled later seems to be the intended behavior of quickjs. Quickjs just calls the handler for every promise that is rejected. The `is_handled` flag doesn't mean that the promise has no handlers but that it has not been handled yet. 

I am not sure that changing the behavior of this handler is the proper thing to do for a rquickjs which should be a wrapper around quickjs. However, I do think that your use case for handling uncaught promise rejections is an important one.

One way we could solve this is by patching quickjs to add a function which checks if a promise has attached rejection handlers. Then we could keep the behavior of rejection handler the same and you could implement your use case by checking if the promise has any handlers. This would also allow catching uncaught promises the moment they are rejected.

richarddd:
> One way we could solve this is by patching quickjs to add a function which checks if a promise has attached rejection handlers. Then we could keep the behavior of rejection handler the same and you could implement your use case by checking if the promise has any handlers. This would also allow catching uncaught promises the moment they are rejected.

I think this is a way better approach. I could investigate how to to check if the promise has any handlers and then call the callback from a patch in quickjs.c. However, i'm not a very good C programmer so i might take i while for me to figure this out :)

PR #120, opened 2023-03-19
Use `Function.prototype` as a proto for Rust functions

katyo:
Fixes issue #59
Alternative for pr #118

### DIFF
--- a/core/src/value/function/ffi.rs
+++ b/core/src/value/function/ffi.rs
@@ -1,5 +1,5 @@
 use super::Input;
-use crate::{qjs, ClassId, Ctx, Object, Result, Value};
+use crate::{qjs, ClassId, Ctx, Result, Value};
 use std::{ops::Deref, panic::AssertUnwindSafe, ptr};
 
 static FUNC_CLASS_ID: ClassId = ClassId::new();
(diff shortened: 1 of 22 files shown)


core/src/value/function/ffi.rs
```diff
@@ -30,7 +30,13 @@ impl<'js> JsFunction<'js> {
     }
 
     pub unsafe fn into_js_value(self, ctx: Ctx<'_>) -> qjs::JSValue {
-        let obj = qjs::JS_NewObjectClass(ctx.ctx, Self::class_id() as _);
+        let proto = ctx
+            .globals()
+            .get::<_, Object>("Function")
+            .unwrap()
+            .get::<_, Object>("prototype")
+            .unwrap();
```
katyo, line 38: Seems It will be better cache `Function.prototype` within context instead lookup it each time.

DelSkayn: Yes, that was what I was thinking to, not caching the object could in fact result in panics. The `Function` field on the global object is also writable. So a different value could be assigned to the `Function` field which would cause this to panic.

DelSkayn:
I am somewhat ambivalent between my approach and this one. The quickjs API seems to nudge towards my approach as it uses the api methods specifically for creating functions. But this one doesn't require and extra object.

While looking around the source code of quickjs I noticed a lot of code specifically handling C functions as created by `JS_NewCFunction*`. I wonder if there might be some differences between how function created by those methods and function created by setting the call field on the quickjs class object. I just haven't had enough time to actually check.

DelSkayn:
I was aware of the prototype object being stored on the context struct.
The only problem I have with patching the source to access it is that that would mean the library not relies on the internal functioning of quickjs which the library normally doesn't expose.
We off course already have a bunch a functionality which patches the source to access internal quickjs stuff but all these are hidden behind features.
If we merge this PR we will always have to patch the quickjs source to build the library something I kinda wanted to avoid.
This is also I explicitly tried to use the function creation API of quickjs in my PR, as this method uses exposed API and thus probably more stable with respect to changes to quickjs over time.

DelSkayn (approved):
Now that I think about it some more, we could probably always fall back on my approach if upstream breaks this one.

katyo:
@DelSkayn,
As I mentioned above `JSCFunction` (and some other like things) internally represented via class with `Function.prototype` as a proto. Currently rquickjs also implements rust functions as a class. The problem in a prototype only. Why not just use correct prototype as `JS_CFunction` does?

Original QuickJS library is not perfect in many aspects. We already added many patches to get it work as we want. Another one (quite simple) fix  does not looks like a big problem for me.

DelSkayn:
> 

Yes agreed, I was more worried about possible future breaking changes in upstream, but as I mentioned in my approval, we can always fallback to my approach if that ever where to happen.

katyo:
@DelSkayn Ok, I will merge it. Thanks a lot for review!

PR #121, opened 2023-03-20
Wrap raw pointers with `NonNull`

katyo:
Also turn exception handling helpers into `Ctx` functions and add missing unsafe specifier to some functions.


DelSkayn (approved):
Looks great! nothing to add.

PR #136, opened 2023-04-09
implement patch manually without using external software

### LINKED ISSUE #88: Unable to build on windows
Hi all,
  Keep getting this error when trying to build for windows, I've been unable to figure out what files I'm missing. Any help is appreciated.

```
Compiling rquickjs-sys v0.1.7
error: failed to run custom build command for `rquickjs-sys v0.1.7`

Caused by:
  process didn't exit successfully: `C:\Users\Zac Reid\Documents\GitHub\surrealdb-test\target\debug\build\rquickjs-sys-136287e45d48a699\build-script-build` (exit code: 101)
  --- stdout
  cargo:rerun-if-changed=build.rs
  cargo:rerun-if-env-changed=CARGO_FEATURE_EXPORTS
  cargo:rerun-if-env-changed=CARGO_FEATURE_BINDGEN
  cargo:rerun-if-env-changed=CARGO_FEATURE_UPDATE_BINDINGS
  cargo:rerun-if-env-changed=CARGO_FEATURE_DUMP_BYTECODE
  cargo:rerun-if-env-changed=CARGO_FEATURE_DUMP_GC
  cargo:rerun-if-env-changed=CARGO_FEATURE_DUMP_GC_FREE
  cargo:rerun-if-env-changed=CARGO_FEATURE_DUMP_FREE
  cargo:rerun-if-env-changed=CARGO_FEATURE_DUMP_LEAKS
  cargo:rerun-if-env-changed=CARGO_FEATURE_DUMP_MEM
  cargo:rerun-if-env-changed=CARGO_FEATURE_DUMP_OBJECTS
  cargo:rerun-if-env-changed=CARGO_FEATURE_DUMP_ATOMS
  cargo:rerun-if-env-changed=CARGO_FEATURE_DUMP_SHAPES
  cargo:rerun-if-env-changed=CARGO_FEATURE_DUMP_MODULE_RESOLVE
  cargo:rerun-if-env-changed=CARGO_FEATURE_DUMP_PROMISE
  cargo:rerun-if-env-changed=CARGO_FEATURE_DUMP_READ_OBJECT

  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: NotFound, message: "program not found" }', C:\Users\Zac Reid\.cargo\registry\src\github.com-1ecc6299db9ec823\rquickjs-sys-0.1.7\build.rs:136:10
  stack backtrace:
     0: std::panicking::begin_panic_handler
               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\std\src\panicking.rs:584
     1: core::panicking::panic_fmt
               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\core\src\panicking.rs:142
     2: core::result::unwrap_failed
               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\core\src\result.rs:1805
     3: enum$<core::result::Result<std::process::Child,std::io::error::Error> >::unwrap<std::process::Child,std::io::error::Error>
               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f\library\core\src\result.rs:1098
     4: build_script_build::patch<ref$<std::path::Path>,std::path::PathBuf>
               at .\build.rs:131
     5: build_script_build::main
               at .\build.rs:98
     6: core::ops::function::FnOnce::call_once<void (*)(),tuple$<> >
               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f\library\core\src\ops\function.rs:248
```

stevefan1999-personal:
This PR attempts to manually do patches without using GNU Patch. Despite the algorithm is quite primitive and had some questionable assumptions, at least it does compile in Windows natively now. The edge case is that I don't know how to determine the correct file to patch -- so I just assumed the `a/{filename}` and `b/{filename}` format.

Fixes #88. Since I made this in Windows and I have to compile it by myself. So I can't possibly made this PR possible if there are any errors. No need to test for Windows therefore.

### DIFF
--- a/sys/Cargo.toml
+++ b/sys/Cargo.toml
@@ -12,6 +12,8 @@ repository = "https://github.com/DelSkayn/rquickjs.git"
 
 [build-dependencies]
 cc = "1"
+diffy = "0.3.0"
+newline-converter = "0.3.0"
 
 [build-dependencies.bindgen-rs]
 package = "bindgen"
--- a/sys/build.rs
+++ b/sys/build.rs
@@ -1,9 +1,7 @@
-use std::{
-    env, fs,
-    io::Write,
-    path::Path,
-    process::{Command, Stdio},
-};
+use std::{env, fs, path::Path};
+
+use diffy::{apply, Patch};
+use newline_converter::dos2unix;
 
 fn main() {
     #[cfg(feature = "logging")]
@@ -129,21 +127,55 @@ fn feature_to_define(name: impl AsRef<str>) -> String {
 }
 
 fn patch<D: AsRef<Path>, P: AsRef<Path>>(out_dir: D, patch: P) {
-    let mut child = Command::new("patch")
-        .arg("-p1")
-        .stdin(Stdio::piped())
-        .current_dir(out_dir)
-        .spawn()
-        .expect("Unable to execute patch, you may need to install it: {}");
-    println!("Appliyng patch {}", patch.as_ref().display());
-    {
-        let patch = fs::read(patch).expect("Unable to read patch");
+    struct Patches<'a>(&'a str);
 
-        let stdin = child.stdin.as_mut().unwrap();
-        stdin.write_all(&patch).expect("Unable to apply patch");
+    // A backtracking attempt to find a valid diff when there are multiple patches in one file (also known as patchset)
+    impl<'a> Iterator for Patches<'a> {
+        type Item = Patch<'a, str>;
+        fn next(&mut self) -> Option<Self::Item> {
+            let mut range = self.0.len();
+            loop {
+                match diffy::Patch::from_str(&self.0[..range]) {
+                    Ok(x) if x.hunks().is_empty() => break None,
+                    Err(_) if range < 1 => break None,
+                    Ok(x) => {
+                        self.0 = &self.0[range..];
+                        break Some(x);
+                    }
+                    Err(_) => range -= 1,
+                }
+            }
+        }
     }
 
-    child.wait_with_output().expect("Unable to apply patch");
+    println!("Appliyng patch {}", patch.as_ref().display());
+    {
+        let out_dir = out_dir.as_ref();
+        let patch = fs::read_to_string(patch).expect("Unable to read patch");
+        let patch = dos2unix(&patch);
+        for patch in Patches(&patch) {
+            let original = patch
+                .original()
+                .and_then(|x| x.split_once('/').map(|(_, b)| b))
+                .expect("Cannot find original file name");
+            let modified = patch
+                .modified()
+                .and_then(|x| x.split_once('/').map(|(_, b)| b))
+                .expect("Cannot find modified file name");
+
+            // for now
+            assert_eq!(original, modified);
+
+            let original =
+                fs::read_to_string(out_dir.join(original)).expect("Unable to read original file");
+            let original = dos2unix(&original);
+            match apply(&original, &patch) {
+                Ok(patched) => fs::write(out_dir.join(modified), patched)
+                    .expect("Unable to write the patched content"),
+                Err(e) => eprintln!("Unable to write the patched content: {}", e),
+            }
+        }
+    }
 }
 
 #[cfg(not(feature = "bindgen"))]
--- a/sys/build.rs
+++ b/sys/build.rs
@@ -157,10 +157,12 @@ fn patch<D: AsRef<Path>, P: AsRef<Path>>(out_dir: D, patch: P) {
             let original = patch
                 .original()
                 .and_then(|x| x.split_once('/').map(|(_, b)| b))
+                .or(patch.original())
                 .expect("Cannot find original file name");
             let modified = patch
                 .modified()
                 .and_then(|x| x.split_once('/').map(|(_, b)| b))
+                .or(patch.modified())
                 .expect("Cannot find modified file name");
 
             // for now


sys/build.rs
```diff
@@ -129,21 +127,57 @@ fn feature_to_define(name: impl AsRef<str>) -> String {
...
+                .or(patch.modified())
+                .expect("Cannot find modified file name");
+
+            // for now
+            assert_eq!(original, modified);
+
+            let original =
+                fs::read_to_string(out_dir.join(original)).expect("Unable to read original file");
+            let original = dos2unix(&original);
+            match apply(&original, &patch) {
```
katyo, line 174: Does it support fuzzy patch applying?
Could you test build with `parallel` feature on windows?

sys/Cargo.toml
```diff
@@ -12,6 +12,8 @@ repository = "https://github.com/DelSkayn/rquickjs.git"
 
 [build-dependencies]
 cc = "1"
+diffy = "0.3.0"
+newline-converter = "0.3.0"
```
katyo, line 16: Avoid patch number in dependency version.

sys/build.rs
```diff
@@ -129,21 +127,57 @@ fn feature_to_define(name: impl AsRef<str>) -> String {
...
+                .or(patch.original())
+                .expect("Cannot find original file name");
+            let modified = patch
+                .modified()
+                .and_then(|x| x.split_once('/').map(|(_, b)| b))
+                .or(patch.modified())
+                .expect("Cannot find modified file name");
+
+            // for now
+            assert_eq!(original, modified);
```
katyo, line 169: Remove that.

sys/build.rs
```diff
@@ -129,21 +127,57 @@ fn feature_to_define(name: impl AsRef<str>) -> String {
...
+
     println!("Appliyng patch {}", patch.as_ref().display());
     {
-        let patch = fs::read(patch).expect("Unable to read patch");
-
-        let stdin = child.stdin.as_mut().unwrap();
-        stdin.write_all(&patch).expect("Unable to apply patch");
+        let out_dir = out_dir.as_ref();
+        let patch = fs::read_to_string(patch).expect("Unable to read patch");
+        let patch = dos2unix(&patch);
```
katyo, line 155: Does we actually need that for both patch and source?
Does we need it at all?
I mean we may simply fix our patches if you have problem with it.

katyo (changes requested):

sys/build.rs
stevefan1999-personal: How do you determine which file to patch?

sys/build.rs
stevefan1999-personal: Not sure about fuzzy patching, but I am building it with parallel and futures

sys/build.rs
katyo: @stevefan1999-personal Course, I about line-ending conversion.

sys/build.rs
katyo: The fact is that MSVC patches changes line numbers and may interfere applying other patches.

sys/build.rs
stevefan1999-personal: > @stevefan1999-personal Course, I about line-ending conversion.

I'm not sure what you mean

katyo: Why we need the following:
```rust
let patch = dos2unix(&patch);
```
and
```rust
let original = dos2unix(&original);
```

stevefan1999-personal: I have CRLF in Windows and diffy does not support CRLF yet. There is no allocation cost if you have LF.

stevefan1999-personal: I know this could turn political quickly, but since QuickJS is written with GCC in mind, and is written by Fabrice Bellard, a legendary open source developer, I'm pretty sure he wrote this while only running on Linux, and Linux is LF by default. 

Yet on Git and Windows, you have automatic LR to CRLF conversion. It is up to the user to turn it off, but it seems like it is a viral option, and I believe this would cause even more unwanted trouble to end-users. 

I'm not aware if `autocrlf` can be turned off per Git repository or can be disabled for Git submodules but anyway, it is almost cost-free if you run this on Unix-like operating system. We should use do this conversion to "normalize" anyway.

rivy: @stevefan1999-personal 

Using this *.gitattributes* will, by default, force text files to use LF EOLs, with CRLF exceptions where required for WinOS or MSVC.

```
# .gitattributes (git attributes config file)

# default; use LF EOLs for text files
* text=auto eol=lf

# CRLF required; force required CRLF EOLs for WinOS BAT/CMD and MSVC SLN/VCPROJ/VCXPROJ files
*.[bB][aA][tT]                      text eol=crlf
*.[cC][mM][dD]                      text eol=crlf
*.[sS][lL][nN]                      text eol=crlf
*.[Vv][Cc][Pp][Rr][Oo][Jj]          text eol=crlf
*.[Vv][Cc][Pp][Rr][Oo][Jj].*        text eol=crlf
*.[Vv][Cc][Xx][Pp][Rr][Oo][Jj]      text eol=crlf
*.[Vv][Cc][Xx][Pp][Rr][Oo][Jj].*    text eol=crlf
```

DelSkayn:
Closing this PR as we no longer need to apply patches to QuickJS.

PR #148, opened 2023-05-24
Change error handling

### LINKED ISSUE #138: rquickjs::Error::Exception should preserve the original value
In JavaScript, while uncommon, it's possible to throw arbitrary values, not just `Error` objects and some libraries actually make use of it.

Right now `rquickjs` doesn't allow to retrieve this non-`Error` value and fails with conversion errors instead, as the `Exception` variant can't represent anything but error objects:

```rust
Exception {
    message: StdString,
    file: StdString,
    line: i32,
    stack: StdString,
},
```

Example:

```rust
fn main() {
    let runtime = rquickjs::Runtime::new().unwrap();
    let context = rquickjs::Context::full(&runtime).unwrap();
    context.with(|ctx| ctx.eval::<(), _>("throw 42;")).unwrap();
}
```

results in the following conversion error, with no way to retrieve the original error value:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: FromJs { from: "int", to: "object", message: None }', /home/rreverser/.cargo/git/checkouts/rquickjs-da202ee7abfef64e/29c5454/core/src/result.rs:478:41
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

For compatibility with arbitrary JS code, it should be possible to retrieve the original `rquickjs::Value` from the exception. As far as I can tell, QuickJS provides this functionality, and the conversion is imposed only at the rquickjs level.

DelSkayn:
This PR restructures the way we handle Javascript exceptions.

Previously we handled all errors thrown by quickjs by transforming them into `Error::Exception{ ... }`. 
`Error::Exception` expected an exception to be an instance of the javascript class `Error` and then retrieved fields from the error object and stored them as rust values.
If the throw exception was not an instance of `Error` the library actually panicked before #141.

This approach has a rather large problem, it is impossible to retrieve the original thrown exception value (see #138).

The most straight forward way to solve this would be to store the original value into the `Error::Exception` variant. 
Using this approach would mean that `Error` would need to have a lifetime to ensure the possible value in exception is handled correctly. 
When experimenting with this approach I found it to be a rather cumbersome change. A lot of the errors raised by the library contain no quickjs values yet they still need an lifetime just for the the `Exception` variant.

Instead I opted to remove any data from the `Exception` variant and have it be a marker that an javascript error was thrown.
If one wants to retrieve the thrown value this can be done with `Ctx::catch`. This method returns the last thrown error.
## example
```rust
if let Err(Error::Exception) = some_func(){
    let error_val = ctx.catch();
    handle_error_val(error_val)
}
```
For convenience I also introduced `CaughtError` which is an error type which has the possible exception value stored as part of the error. Defined as:
```rust
enum CaughtError<'js>{
    Exception(Exception<'js>),
    Value(Value<'js>),
    Error(Error)
}
```
`Exception<'js>` is a wrapper around object for instances of `Error` which implements `Display` for printing the error.
If the thrown javascript error is not an instance of `Error` it is stored in `CaughtError::Value`.
I also introduced various methods to easily convert between `Error` and `CaughtError`. The above example can also be written as:
```rust
if let Err(CaughtError::Exception(error_val)) = some_func().catch(ctx){
    handle_error_val(error_val)
}
```
## Advantages
- No lifetime on `Error`
- Easily supports the most common case of exceptions where they are just immediately returned.
- Allows one to retrieve the error value regardless of value type.
- Makes no assumptions about how error values should be handled.
## Disadvantages
- More complex error handling
- `Error` no longer prints the quickjs error information, users must transform `Error` into `CaughtError` if they want the quickjs error printed.
- Introduces state for errors which could possibly be overwritten if not handled properly.

Merging this would fix #138.


RReverser:
> Using this approach would mean that `Error` would need to have a lifetime to ensure the possible value in exception is handled correctly.

Couldn't it store Exception as a Persistent member instead?

DelSkayn:
After using this version of the library I found that this way of handling errors has some inconveniences here and there but I think those can be addressed. 
I think this way of handling errors is a good basis and allows us to implement more convenient abstraction on top of it in the future.
So I am merging it.

RReverser:
Thanks!

PR #171, opened 2023-06-29
Fix implementation of the `Debug` trait of `Value`

MaxXSoft:
```rs
if let Err(Error::Exception) = ctx.eval::<(), _>("throw 3") {
  println!("{:?}", ctx.catch());
}
```

The result of the above code is `Int'('3')'` but not `Int(3)`.

Similarly, the result of the following code

```rs
if let Err(Error::Exception) = ctx.eval::<(), _>(r#"throw "3""#) {
  println!("{:?}", ctx.catch());
}
```

is `String"(\""Ok("3")"\")"` but not `String(Ok("3"))`.

This PR fixes above issues.


DelSkayn:
Thanks for the PR, merging!

PR #175, opened 2023-07-09
Add value reference count getter

makindotcc:


richarddd:
@makindotcc thanks for the PR. Can you please adds some tests?

makindotcc:
Hello @richarddd. After merge pr is no longer compiling due to this commit:
https://github.com/DelSkayn/rquickjs/commit/7988333d3d915ada06318134b6c130b849b6a8dc#diff-30624fb4011d0b71d6e715a9cee4c3d70c8645995ac06423a16af30768ef5b1eL84-L88

I'm not sure how to handle it, since JSRefCountHeader is no longer in quickjs.h, but in quickjs.c. I should also check ``JS_VALUE_HAS_REF_COUNT`` before grabbing refcount.
I think correct replacement for my pr would be api to ``JSClassDef.finalizer``, but I can't provide it now, sorry!
Closing pr, because I think this is objectively wrong.

PR #184, opened 2023-07-22
deps: Missing convert_case from recent updates.

waywardmonkeys:
Looks like you missed this when landing recent changes.


DelSkayn:
Good catch!

I missed a change when rebasing.

However I had already fixed the missing dependency locally when working on some other changes so I am closing this PR.

PR #201, opened 2023-08-16
Fix spawner ordering

richarddd:
I noticed some timing issues when using `ctx.spawn`. Tracked down the issue to lie in `swap_remove` where the order of how futures are spawned will be modified.

I modified the logic to collect and remove all completed futures in a single pass and then remove them in reverse order to avoid index shifting as much as possible.

A linked list rather than a Vec is O(1) for removal during iteration and insertion at end but will involve pointer following rather than array iteration that might not be ideal.

It's also better for performance if we pull and remove all completed tasks.


DelSkayn:
Looks good to me!

Will merge when the CI completes!

PR #206, opened 2023-09-01
clippy: Fix filter-map-bool-then warnings.

waywardmonkeys:
Use `filter` then `map, rather than `filter_map` with a `bool::then` inside of it.


waywardmonkeys:
This is a new check in Rust 1.72.

DelSkayn:
Looks good.

Thx!

PR #222, opened 2023-10-03
Add column number to error

richarddd:
Currently quickjs does not keep track of column numbers where an error occurs. Users running minified JavaScript can not debug their code if there is column number present as line number will always be 1.

This PR applies the line number patch from the quickjs repo.


DelSkayn (approved):
LGTM

Issue #224, opened 2023-10-06
Crash due to memory leak & undefined behaviour in spawn poller when spawning nested tasks


richarddd:
Upon further investigation of https://github.com/DelSkayn/rquickjs/issues/223 i discovered a memory leak in https://github.com/DelSkayn/rquickjs/blob/fb7d128d512d9f9a11024c8594e7486e3d350ae6/core/src/runtime/spawner.rs#L70 resulting in a crash:

```
Assertion failed: (list_empty(&rt->gc_obj_list)), function JS_FreeRuntime, file quickjs.c
[1]    39846 abort 
```

Consider the following code:

```rust
let mut did_complete = false;
self.0.futures.retain_mut(|f| {
    let ready = f.as_mut().poll(cx).is_ready();
    did_complete = did_complete || ready;
    !ready
});
```

If the `f.as_mut().poll(cx)` contains a new `ctx.spawn` it will modify the `futures` vec causing undefined behaviour.

That's why my original PR https://github.com/DelSkayn/rquickjs/pull/201/commits/982c4640a8a054704a8f98f5efe8975c75fd0d15 correcting the spawn ordering used a vec to collect the indexes and then remove them to avoid concurrent modification of the futures vec:

```rust
fn poll(mut self: Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> Poll<Self::Output> {
  if self.0.futures.is_empty() {
      return Poll::Ready(false);
  }

  let mut completed_indices = Vec::new();

  // Iterate over the futures and check their completion status.
  for (i, f) in self.0.futures.iter_mut().enumerate() {
      if f.as_mut().poll(cx).is_ready() {
          completed_indices.push(i);
      }
  }

  // Remove the completed futures in reverse order to avoid index shifting.
  for &idx in completed_indices.iter().rev() {
      self.0.futures.remove(idx);
  }

  if !completed_indices.is_empty() {
      Poll::Ready(true)
  } else {
      Poll::Pending
  }
}
```

PR #225, opened 2023-10-06
Revert "Optimize poll loop slightly"

richarddd:
This reverts commit ad762253516e7230f14a9b6984b13250b811a5d9.

Fixes https://github.com/DelSkayn/rquickjs/issues/224

### DIFF
--- a/core/src/runtime/spawner.rs
+++ b/core/src/runtime/spawner.rs
@@ -65,14 +65,21 @@ impl<'a, 'js> Future for SpawnFuture<'a, 'js> {
             return Poll::Ready(false);
         }
 
-        let mut did_complete = false;
-        self.0.futures.retain_mut(|f| {
-            let ready = f.as_mut().poll(cx).is_ready();
-            did_complete = did_complete || ready;
-            !ready
-        });
+        let mut completed_indices = Vec::new();
 
-        if did_complete {
+        // Iterate over the futures and check their completion status.
+        for (i, f) in self.0.futures.iter_mut().enumerate() {
+            if let Poll::Ready(_) = f.as_mut().poll(cx) {
+                completed_indices.push(i);
+            }
+        }
+
+        // Remove the completed futures in reverse order to avoid index shifting.
+        for &idx in completed_indices.iter().rev() {
+            self.0.futures.remove(idx);
+        }
+
+        if !completed_indices.is_empty() {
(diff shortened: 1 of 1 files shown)


DelSkayn:
Good catch! I didn't think about new futures being added while polling, however, won't the reverted code have the same problem as the current? 

I believe the iterator returned by `iter_mut` will also be invalidated if a value is pushed into the vector, as they consist of just two pointers pointing to the memory of the vector.

The ultimate problem is that we have aliasing mutable references which will always be unsound.

I think the only proper solution is to implement the schedular with a linked list.

richarddd:
> Good catch! I didn't think about new futures being added while polling, however, won't the reverted code have the same problem as the current?

Nope, because we are not mutating the array from inside the iterator. We are simply collecting indexes. 

> I believe the iterator returned by `iter_mut` will also be invalidated if a value is pushed into the vector, as they consist of just two pointers pointing to the memory of the vector.
> 
> The ultimate problem is that we have aliasing mutable references which will always be unsound.
> 
> I think the only proper solution is to implement the schedular with a linked list.

A double linked list might be an alternative but i suspect pointer following might have slight performance implications for lots of spawns.

DelSkayn:
> Nope, because we are not mutating the array from inside the iterator. We are simply collecting indexes.

I disagree, the for loop holds onto a iterator to the futures vector which is just a pointer to the memory of the vector. Then inside the for loop `poll` gets called which can push a new value into the futures vec. If that happens the futures vec memory might need to be reallocated, which can cause the memory which the iterator pointer is pointing to, to become invalid resulting, again, in a segfault.

We could possibly change to using indexes in the array, but I believe, because we are supposed to have exclusive access to the futures vec the compiler could decide to still optimize the array index away to again using pointers.

> A double linked list might be an alternative but i suspect pointer following might have slight performance implications for lots of spawns.

If it is done right the schedular wouldn't even have to follow pointers that often. I have been thinking about implementing the schedular with two linked lists such that whenever a future calls `wake` on the waker, that waker moves the future into the _should be polled_ list from the _sleeping_ list. That way the shedular knows which futures to poll again, avoiding the possible quadratic complexity the schedular currently has.

richarddd:
Closing in favor of https://github.com/DelSkayn/rquickjs/pull/228

PR #245, opened 2024-01-10
add support for wasm32-wasi

mmoskal:
This makes build.rs download wasi-sys if building for wasm32-wasi target. The logic is copied from https://github.com/bytecodealliance/javy/blob/main/crates/quickjs-wasm-sys/build.rs

I was unable to get the bindgen to work - it would not output functions (it would output types and constants correctly though). In the end I copied the bindgen file from i686-msvc. It works OK.


sys/build.rs
```diff
@@ -168,7 +285,7 @@ where
         .unwrap_or(false)
     {
         println!(
-            "cargo:warning=rquickjs probably doesn't ship bindings for platform `{}`. try the `bindgen` feature instead.",
+            "cargo:warning=rquickjs x probably doesn't ship bindings for platform `{}`. try the `bindgen` feature instead.",
```
DelSkayn, line 288: This `x` here shouldn't be there.
```suggestion
            "cargo:warning=rquickjs probably doesn't ship bindings for platform `{}`. try the `bindgen` feature instead.",
```

DelSkayn (changes requested):
Looks good apart from one minor change in a println.

DelSkayn:
Thanks for the PR!

I have also been experimenting with adding support for wasi. I thought I managed at one point to get bindgen working, so I will do some testing with that before merging this.

sys/build.rs
```diff
@@ -101,16 +184,50 @@ fn main() {
...
+        out_dir,
+        out_dir.join("quickjs.bind.h"),
+        &defines,
+        add_cflags,
+    );
 
     let mut builder = cc::Build::new();
     builder
         .extra_warnings(false)
+        .flag("-Wno-implicit-const-int-float-conversion")
```
DelSkayn, line 217: This should probably be `flag_if_supported`. This flag doesn't exist on windows and cause the build to fail.

sys/build.rs
mmoskal: whoops, fixed

sys/build.rs
mmoskal: fixed

dejang:
Really appreciate the work done here, do you know how we could make this work for the web ? I'm really curios about that. I' m currently working on a project where I'm in need of an isolated environment that is not an iframe or worker and QuickJS fits perfectly.

DelSkayn:
@dejang the problem with compiling quickjs for web wasm is that there are currently some problems with packaging c libraries which require libc. The default wasm target used by rust is `wasm-unknown-unknown` which doesn't have a libc available. We could instead add support for `wasm-emscripten` but this requires a lot of code just to pull in the sdk or check for it's availability for a platform which isn't well supported by rust tools like `wasm-bindgen`. 

For more reference see rustwasm/team#291

dejang:
Thank you @DelSkayn, that actually makes a lot of sense.

PR #267, opened 2024-02-20
Add name to constructor when registering constructor

richarddd:
Name property is not set when creating constructors. This PR adds the name from the class to the constructor so the following will work for classes defined in Rust.

```javascript
new MyClass().constructor.name == "MyClass"
```


DelSkayn (approved):
Looks good thanks!

PR #275, opened 2024-03-05
use global wasi-sdk if WASI_SDK env is set

Gentle:
this patch allows to use an external wasi sdk by setting the environment variable `WASI_SDK` to the path to be used.

My usecase is that I have a wasi-sdk for linux-aarch64 installed in /opt/wasi-sdk and there is currently no binary package to install for that architecture


DelSkayn:
Looks good! Thanks for the PR!

Issue #266, opened 2024-02-18
Add support to "aarch64-linux-android"


yhyadev:
It's just like aarch64-unknown-linux-gnu I think..

yhyadev:
You can just use the bindgen feature

PR #291, opened 2024-04-18
Add bindings for aarch64-linux-android

ahaoboy:
Although you can use the bindgen feature https://github.com/DelSkayn/rquickjs/issues/266 , bindgen_rs fails to compile in some scenarios, adding aarch64-linux-android will make compiling faster and easier.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ members = [
     "macro",
     "examples/native-module",
     "examples/module-loader",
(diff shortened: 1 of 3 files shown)


ahaoboy:
It seems not as easy as imagined...

Sytten:
I would suggest looking at cross-rs (https://github.com/cross-rs/cross), it simplifies a lot the setup.

PR #297, opened 2024-04-25
Remove #[non_exhaustive] from EvalOptions

richarddd:
I saw this was added in 0.6.0 update

Right now we can't eval as a promise but apply non-strict or non-global for instance.


richarddd:
I was stupid, i might as well just mutate it! 🤦‍♂️

PR #300, opened 2024-04-29
Fix intrinsic loading order

surma:
It is very sensible that since v0.6, the `BaseObject` intrinsic is always loaded. However, in a custom setup, the `BaseObject` intrinsic is loaded _after_ the custom intrinsics, which means all the custom intrinsics cannot assume `BaseObject`. If `BaseObject` is added to the list of custom intrinsics, `BaseObject` gets loaded twice, which causes a SIGABORT.

Additionally, `All` still contains `Base`, which similarly causes `SIGABORT`.

This PR loads `Base` _before_ the custom intrinsics. It also removes `Base` from `All` (and adds some tests for this).


DelSkayn (approved):

DelSkayn:
Thanks for the PR! Looks good!

PR #301, opened 2024-04-29
Regenerate WASI bindings

surma:
The bindings file for the `wasm32-wasi` target wasn’t regenerated. This fixes that.

(For others finding this PR: The workaround in the meantime is to enable the `bindgen` feature on the `rquickjs` crate).


DelSkayn:
I added support for automated binding generation in CI for wasm in #298 so the bindings are already updated and will be automatically updated in the future. So this PR is no longer needed.

PR #305, opened 2024-04-30
Use up to date version config in build.rs

saulecabrera:
This commit fixes the QuickJS version used in the build file, which should match the version defined in [quickjs/VERSION](https://github.com/bellard/quickjs/blob/3b45d155c77bbdfe9177b1e03db830d2aff0b2a8/VERSION)


DelSkayn:
Thanks!

PR #312, opened 2024-05-13
Make sure promise polls spawner

richarddd:
Promise converted into a future will now advance spawned tasks

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ members = [
 [workspace.dependencies]
 rquickjs-core = { version = "0.8.1", path = "core" }
 rquickjs-macro = { version = "0.8.1", path = "macro" }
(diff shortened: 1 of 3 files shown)


richarddd:
@DelSkayn can we merge this if put behind a feature flag?

richarddd:
@DelSkayn this is now gated behind `promise_poll_spawner`. Would that be OK?

DelSkayn:
There is currently still an issue on this PR. It is possible for a task to become unblocked while it is polling, which would allow it to poll itself, finish and then free it's own task.

richarddd:
> There is currently still an issue on this PR. It is possible for a task to become unblocked while it is polling, which would allow it to poll itself, finish and then free it's own task.

Can we track this somehow? Or add primitives to prevent this?

PR #313, opened 2024-05-14
Expose methods on spawner in the API.

DelSkayn:
Adds the poll_schedular and listen_spawn method.

### DIFF
--- a/core/src/context/ctx.rs
+++ b/core/src/context/ctx.rs
@@ -1,5 +1,3 @@
-#[cfg(feature = "futures")]
-use std::future::Future;
 use std::{
     ffi::{CStr, CString},
     fs,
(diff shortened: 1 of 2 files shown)


DelSkayn:
Hmmm, I am still having some problems finding a good API here.

This PR has the problem that exposing poll and such introduces the ability to poll recursively: polling the scheduler from within a future currently being polled in the scheduler. The scheduler wasn't really designed with that in mind. I can add support for it but it is still hacky and I am not sure I want to expose this type of API.

richarddd:
> Hmmm, I am still having some problems finding a good API here.
> 
> This PR has the problem that exposing poll and such introduces the ability to poll recursively: polling the scheduler from within a future currently being polled in the scheduler. The scheduler wasn't really designed with that in mind. I can add support for it but it is still hacky and I am not sure I want to expose this type of API.

Thanks for the status update, I really appreciate the effort. Does this also mean that https://github.com/DelSkayn/rquickjs/pull/312 is obsolete? All current blockers goes away if promises also drove spawned tasks forward.

PR #317, opened 2024-05-24
Add Null support for function

Sytten:
Missing Null implementations

### DIFF
--- a/core/src/value/function.rs
+++ b/core/src/value/function.rs
@@ -389,6 +389,28 @@ mod test {
     }
 
     #[test]
+    fn call_js_fn_with_null_args_and_return() {
+        let res: Vec<i8> = test_with(|ctx| {
+            let func: Function = ctx
+                .eval(
+                    r#"
+                  (a, b) => {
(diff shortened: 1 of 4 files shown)


Sytten:
The conflict with the `Null` struct is kinda of annoying, we would also need a `Nullable` struct that has a `IntoJs` since `Option` returns `undefined`.

DelSkayn:
I think this type should be called `Nullable` not `Null` since that already conflicts with the null value struct.
Nullable also makes more sense I think cause the type is a wrapper around a value which might be null. 

Also instead of implementing FromParams and IntoArgs maybe it should instead just implement `IntoJs` and `FromJs`. Then they can be used in function params and everywhere else you would need the absence of a value to be null.

Sytten:
Good point, I wasn't sure what we wanted to do with that since we have `Opt` that is `FromParams` only and we already have the `Null` struct so you can do `Either<Value, Null>` already. Do you want me to remove `function::Null` since it's already in the codebase (I didn't create that struct, it must be from an earlier iteration)?

PR #319, opened 2024-05-24
New util crate

Sytten:
This is a new util crate that is meant as a place for the community to contribute utilities that are not covered by quickjs natively. This first iteration adds:

- Console (~80% on spec)
- UrlSearchParams (100% on spec)

### DIFF
--- a/util/src/console/formatter.rs
+++ b/util/src/console/formatter.rs
@@ -1,0 +1,383 @@
+use std::fmt::Write;
+
+use rquickjs::class::Trace;
+use rquickjs::{Error, Result, Type, Value};
+
+#[derive(Default, Clone, Debug)]
+#[non_exhaustive]
+struct FormatArgs {
+    key: Option<bool>,
+}
+
+impl FormatArgs {
+    pub fn is_key(&self) -> bool {
+        self.key.unwrap_or(false)
+    }
+
+    pub fn with_key(self) -> Self {
+        Self {
+            key: Some(true),
+            ..self
+        }
+    }
+}
+
+/// A formatter for the [`Console`] object
+///
+/// This formatter is used to format values to be printed by the console object.
+///
+/// [`Console`]: crate::console::Console
+#[derive(Clone, Debug, Trace)]
+pub struct Formatter {
+    max_depth: usize,
+}
+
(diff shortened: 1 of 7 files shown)


ahaoboy:
Implementing console in rust is very complicated. The best way may be to reuse deno's js code, which will be very simple at the rust and consistent with deno. Here is an example, I hope it will be helpful.

https://github.com/ahaoboy/rqjs/blob/main/rqjs-ext/src/modules/console.rs

Sytten:
@ahaoboy That sounds like a bad idea considering it is js based and quickjs is really not that fast to run js code. I dont think we need to be 100% there. We have been using this console implementation for about a year.

ahaoboy:
Nice job~


> @ahaoboy That sounds like a bad idea considering it is js based and quickjs is really not that fast to run js code. 


I haven't tested the difference in size and speed between the js and rust code, but the two are not in conflict. Users should be able to choose different console implementations based on their scenario. If they need more style and more comprehensive output, they can choose console_deno(maybe this name). If they need more startup speed and size, then the simple rust implementation will be more suitable.


Based on the test of the following simple code, it seems that Promsie is not supported?
```js
console.log([1, '1'])
console.log(1)
console.log('a')

const a = [1]
a[1] = a;
console.log(a)

console.log('Promise:', Promise.resolve(1))
console.log('Proxy:', new Proxy({ a: 1 }, {}))

```

image

Sytten:
Yeah still missing some things, thats why I said 80% done

ahaoboy:
Consider supporting console.clear function?

util/src/console/formatter.rs
```diff
@@ -0,0 +1,383 @@
...
+            let formatter = Formatter::default();
+            let value = ctx.eval("'test'").unwrap();
+            let mut out = StdString::default();
+            formatter.format(&mut out, value).unwrap();
+            assert_eq!("test", out);
+        })
+    }
+
+    #[test]
+    fn format_int() {
```
gc-victor, line 240: It is testing a boolean case

Sytten:
Closing since I am moving my modules to https://github.com/awslabs/llrt/tree/main/llrt_modules

Issue #318, opened 2024-05-24
Add iterator wrapper


Sytten:
Creating iterators is a pain right now. A basic example:

```rust
 pub fn entries<'js>(&self, ctx: Ctx<'js>, this: This<Self>) -> Result<Object<'js>> {
        let res = Object::new(ctx)?;

        res.set("position", 0usize)?;
        res.set(
            PredefinedAtom::Next,
            Func::from(
                move |ctx: Ctx<'js>, it: This<Object<'js>>| -> Result<Object<'js>> {
                    let position = it.get::<_, usize>("position")?;
                    let res = Object::new(ctx.clone())?;
                    if this.data.len() <= position {
                        res.set(PredefinedAtom::Done, true)?;
                    } else {
                        let (name, value) = &this.data[position];
                        res.set(
                            "value",
                            vec![
                                rquickjs::String::from_str(ctx.clone(), name),
                                rquickjs::String::from_str(ctx, value),
                            ],
                        )?;
                        it.set("position", position + 1)?;
                    }
                    Ok(res)
                },
            ),
        )?;
        Ok(res)
    }
```

I tried creating a wrapper, but I can't get it to play nice:
```rust
/// A trait for converting a Rust
pub trait IntoJsIter<'js> {
    type Item: IntoJs<'js>;

    /// Call the function with the given parameters.
    fn next(&mut self, ctx: Ctx<'js>, position: usize) -> Result<Option<Self::Item>>;
}

#[derive(Clone, Debug, PartialEq, Eq, Hash)]
#[repr(transparent)]
pub struct Iterator<'js>(pub(crate) Object<'js>);

impl<'js> Iterator<'js> {
    /// Create a new function from a Rust function which implements [`IntoJsFunc`].
    pub fn new<T, I>(ctx: Ctx<'js>, mut it: T) -> Result<Self>
    where
        T: IntoJsIter<'js, Item = I> + 'js,
        I: IntoJs<'js>,
    {
        let iterator = Object::new(ctx)?;
        iterator.set("position", 0usize)?;
        iterator.set(
            "next",
            Func::from(
                |ctx: Ctx<'js>, this: This<Object<'js>>| -> Result<Object<'js>> {
                    let position = this.get::<_, usize>("position")?;
                    let res = Object::new(ctx.clone())?;
                    if let Some(value) = it.next(ctx, position)? {
                        res.set("value", value)?;
                        this.set("position", position + 1)?;
                    } else {
[...]

PR #320, opened 2024-05-27
Add iterator

Sytten:
This is my first attempt at https://github.com/DelSkayn/rquickjs/issues/318

Couple of points before reviewing:
- I decided to base the implementation on `ArrayBuffer` since `Iterable` and `Iterator` are not JS types and I didn't feel comfortable adding them to the `Type` enum
- Still unsure if I like the name of the trait `IntoJsIter` since it's not really an `Into` trait like we generally see in rust, but it's similar in spirit to `IntoJsFunc`
- The parameters of `IntoJsIter` are a bit strange with the `position`. I use the `position` in my util crate https://github.com/DelSkayn/rquickjs/pull/319/files#diff-653c6889323b6067b2be818f8ff00f785927ae3beaaeaba4a034feb5ea154d4eR133 since I do some magic with the `This` to access the parent class to avoid copying the values and my function is stateless so I need some indication on where we are. But we might want to remove it from the trait and move it to `IterFn` and `IterFnMut`


Sytten:
Gentle ping @DelSkayn, whenever you have time I just want to avoid it getting stale :)

Sytten:
Need more work

PR #324, opened 2024-06-21
Fixed #[methods(crate = "...")].

FrankHB:
Fixed a typo so `#[methods(crate = "...")]` can work, instead of incorrectly update the configuration of the prefix.


DelSkayn (approved):
Good catch, definitely a missed copy. Thanks!

PR #325, opened 2024-06-21
Add Send + Sync on Persistent

### LINKED ISSUE #323: Make Persistent Send
Right now, if you try to use `Persistent` with the `Parallel` and `futures` flag (`AsyncContext`) you won't be able to do so since `Persistent` is not marked as `Send`.

I am pretty sure this is safe since the `restore` function will take a lock on the runtime implicitly.

I am also confused about the comment:
https://github.com/DelSkayn/rquickjs/blob/304db5d8ebe6b42cfa9d9cf6d11ff7d355bdf2ac/core/src/persistent.rs#L165-L166
I don't understand how that is possible reading the code. Does that mean we should use a manual drop to make sure the `Persistent` is dropped before the runtime or it doesn't matter?

Sytten:
Fixes #323

### DIFF
--- a/core/src/persistent.rs
+++ b/core/src/persistent.rs
@@ -162,8 +162,13 @@ unsafe impl<'js, T> Outlive<'js> for Module<'js, T> {
 /// It is an error (`Error::UnrelatedRuntime`) to restore the `Persistent` in a
 /// context who isn't part of the original `Runtime`.
 ///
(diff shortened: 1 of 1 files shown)


Sytten:
Like was mentioned in the issue it is a bit more complicated then that, it can have a race if dropped when the runtime is not locked.

PR #328, opened 2024-06-28
Update QuickJS dependency

richarddd:
Update QuickJS dependency.

@DelSkayn would you also consider switching to the https://github.com/quickjs-ng/ fork? 

It's more actively maintained, has some great performance improvements and will eventually be merged with QuickJS. 

I've started some work in a branch if you're interested.


DelSkayn:
I was holding of on switching to quickjs-ng because they where going to be merged, I was under the impression that the original would become the base, so I didn't want to have to switch and then in the future switch back. But I am not sure about the timeline of the merge, so if it is still far out we could consider moving to quickjs-ng.

richarddd:
> I was holding of on switching to quickjs-ng because they where going to be merged, I was under the impression that the original would become the base, so I didn't want to have to switch and then in the future switch back. But I am not sure about the timeline of the merge, so if it is still far out we could consider moving to quickjs-ng.

Charlie mentioned fall 24 here:
https://github.com/quickjs-ng/quickjs/issues/374#issuecomment-2051226800

richarddd:
Somehow git checked out the wrong remote. So closing this PR and creating a new one instead!

PR #329, opened 2024-07-02
feat(core, value): add as_bytes_mut() to array buffer and typed array

stephencroberts:
# Overview

There's currently not a way to modify the bytes of an `ArrayBuffer` or a typed array like `Uint8Array`, so this PR adds new `as_bytes_mut()` methods that follows the typical convention in Rust.

### DIFF
--- a/core/src/value/array_buffer.rs
+++ b/core/src/value/array_buffer.rs
@@ -106,6 +106,14 @@ impl<'js> ArrayBuffer<'js> {
     /// Returns `None` if the array is detached.
     pub fn as_bytes(&self) -> Option<&[u8]> {
         let raw = Self::get_raw(self.as_value())?;
+        Some(unsafe { slice::from_raw_parts(raw.ptr.as_ptr(), raw.len) })
+    }
+
+    /// Returns the underlying bytes of the buffer as a mutable slice,
+    ///
+    /// Returns `None` if the array is detached.
+    pub fn as_bytes_mut(&self) -> Option<&mut [u8]> {
+        let raw = Self::get_raw(self.as_value())?;
         Some(unsafe { slice::from_raw_parts_mut(raw.ptr.as_ptr(), raw.len) })
     }
 
--- a/core/src/value/typed_array.rs
+++ b/core/src/value/typed_array.rs
@@ -171,6 +171,14 @@ impl<'js, T> TypedArray<'js, T> {
         Some(unsafe { slice::from_raw_parts(ptr.as_ptr(), len) })
     }
 
+    /// Returns the underlying bytes of the buffer as a mutable slice,
+    ///
+    /// Returns `None` if the array is detached.
+    pub fn as_bytes_mut(&self) -> Option<&mut [u8]> {
+        let (_, len, ptr) = Self::get_raw_bytes(self.as_value())?;
+        Some(unsafe { slice::from_raw_parts_mut(ptr.as_ptr(), len) })
+    }
+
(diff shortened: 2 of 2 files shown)


core/src/value/array_buffer.rs
```diff
@@ -105,6 +105,14 @@ impl<'js> ArrayBuffer<'js> {
...
+        Some(unsafe { slice::from_raw_parts(raw.ptr.as_ptr(), raw.len) })
+    }
+
+    /// Returns the underlying bytes of the buffer as a mutable slice,
+    ///
+    /// Returns `None` if the array is detached.
+    pub fn as_bytes_mut(&self) -> Option<&mut [u8]> {
         let raw = Self::get_raw(self.as_value())?;
         Some(unsafe { slice::from_raw_parts_mut(raw.ptr.as_ptr(), raw.len) })
     }
```
DelSkayn, line 118: This is unsound. It allows you to call `as_bytes_mut` twice on the same object resulting in aliasing mutable references which is undefined behavior. This also cannot be avoided by making the function take a mutable reference to self as you can have two separate `ArrayBuffer` values point to the same slice.

In order to make this sound, there needs to be some kind of RefCell/Mutex like lock to prevent aliasing mutable references.

DelSkayn (changes requested):

core/src/value/array_buffer.rs
stephencroberts: Thanks @DelSkayn that's a great point. I'll take some time to rework it, unless there's already a way to manipulate the bytes directly without copying that I'm missing?

DelSkayn: It is unsafe and also very prone to unsoundness, especially with a scripting language which might call functions in any order this function could quickly be used wrong. Even just creating an aliasing mutable reference to the same object is unsound, even if the mutable reference is never used. 

I am fine with allowing having some unsafe way to access the bytes internally but it should then be via a raw pointer, not a mutable reference. Raw pointers are a lot less prone to accidental unsoundness.

Sytten: In that sense the raw interface that is already present is good enough since I was able to use it for my usecase.

I am just not sure what the worst case scenario is really is if we do `pub unsafe fn as_bytes_mut(&self) -> Option<&mut [u8]>`. We know the slice will be be valid since the GC wont run until `self` is dropped at minimum.
Yes it can be unsound if you have two concurrent access to it (even though that seems to be the expectation for Node), that is why it is marked unsafe and we should add a proper documentation.

This is not a problem unique to rquickjs, deno is in the [same boat here](https://github.com/denoland/deno/issues/16756). They allow `DerefMut` from it, which is even worst.

- V8Slice: https://github.com/denoland/deno_core/blob/665a61d733d4c9cb18aa81c050c135b51e9b51f5/serde_v8/magic/v8slice.rs#L59-L78
- Safety: https://github.com/denoland/deno_core/blob/665a61d733d4c9cb18aa81c050c135b51e9b51f5/serde_v8/magic/v8slice.rs#L180-L186

Their safety comment is "interesting" to say the least.

stephencroberts: So if I update to mark as `unsafe`, do you think that would be acceptable @DelSkayn given @Sytten 's feedback? Or should I just close the MR? We can continue to maintain this code in our own fork of rquickjs if we need to.

Sytten:
@stephencroberts i would accept an unsafe version of that PR, we have maintainership rights now.

PR #330, opened 2024-07-07
feat: add impl IntoJs and FromJs for char

ahaoboy:
In the scenario of passing a single character in js and rust, it is more convenient to use char than String


DelSkayn (approved):

ahaoboy:
ci failed, but it seems to have nothing to do with this pr?

DelSkayn (approved):
Changes look good, thanks!

PR #331, opened 2024-07-10
Update QuickJS

richarddd:
Adds latest changes from upstream


richarddd:
Hi @DelSkayn, i think CI is broken since it targets rust 1.65. Please take a look when you have the time

DelSkayn:
The CI breaking was apparently due to a raise in the MSRV of cc which has since been reverted. I will probably have to raise the MSRV of rquickjs soon when cc bumps it MSRV again.

DelSkayn (approved):

PR #336, opened 2024-07-18
Remove wrong README stuff and add community

### LINKED ISSUE #164: Best place to share utils?
Hi!
I am building a couple of utils that I think could be useful to the community.
The first one is a full `Console` implementation that handles all the types of values.
I think it would be worth sharing I am just wondering if you should create a `rquickjs-extra` repo/crate? Or maybe just examples.
Let me know!

Sytten:
Closes #164


DelSkayn:
Looks good thanks!

PR #338, opened 2024-07-20
add example rquickjs-cli

ahaoboy:
This is a very simple example, but it covers many common use cases.
1. call rust functions in js
2. call js functions in rust
3. run js code in rust

### DIFF
--- a/examples/rquickjs-cli/src/main.rs
+++ b/examples/rquickjs-cli/src/main.rs
@@ -1,0 +1,42 @@
+use std::io::{stdout, Write};
+
+use rquickjs::{Context, Function, Object, Result, Runtime, Value};
+
+fn print(s: String) {
+    println!("{s}");
+}
+
+fn main() -> Result<()> {
+    let rt = Runtime::new()?;
+    let ctx = Context::full(&rt)?;
+
+    ctx.with(|ctx| -> Result<()> {
+        let global = ctx.globals();
+        global.set(
+            "print",
+            Function::new(ctx.clone(), print)?.with_name("print")?,
+        )?;
+        ctx.eval::<(), _>(
+            r#"
+globalThis.console = {
+  log(v){
+    globalThis.print(`${v}`);
+  }
+}
+"#,
+        )?;
+
+        let console: Object = global.get("console")?;
+        let js_log: Function = console.get("log")?;
+        loop {
+            let mut input = String::new();
+            let _ = stdout().write(b"> ")?;
+            stdout().flush()?;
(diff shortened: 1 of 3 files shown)


trevyn:
I found this helpful for quickly testing rquickjs, thanks. :)

examples/rquickjs-cli/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+"#,
+        )?;
+
+        let console: Object = global.get("console")?;
+        let js_log: Function = console.get("log")?;
+        loop {
+            let mut input = String::new();
+            let _ = stdout().write(b"> ")?;
+            stdout().flush()?;
```
DelSkayn, line 34: Why manually lock stdout? `print!("> ")` should be sufficient here.

examples/rquickjs-cli/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+            std::io::stdin().read_line(&mut input)?;
+            let ret = ctx.eval::<Value, _>(input.as_bytes())?;
```
DelSkayn, line 36: This won't report errors properly. You should retrieve the error thrown by QuickJS by calling for example `.catch(&ctx)` on the returned result.

examples/rquickjs-cli/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+            js_log.call::<(Value<'_>,), ()>((ret,))?;
```
DelSkayn, line 37: Same here.

This probably won't throw an error but for an example code it is good to follow best practice.

DelSkayn (commented):
I am fine with adding this example, just have some comments.

examples/rquickjs-cli/src/main.rs
ahaoboy: If flush is not used here to force the output of the buffer content, the prompt will not be display.

examples/rquickjs-cli/src/main.rs
ahaoboy: Is this the right way to use it?
```rust
  match ctx.eval::<Value, _>(input.as_bytes()).catch(&ctx) {
      Ok(ret) => {
          js_log.call::<(Value<'_>,), ()>((ret,))?;
      }
      Err(err) => {
          println!("{err}");
      }
  }
```

image

DelSkayn: Yes that will return the right error.

DelSkayn (approved):

PR #349, opened 2024-08-02
String as_str function

richarddd:
Expose a `as_str` function for JavaScript strings.
Also assume data is valid utf8 on production builds.

### DIFF
--- a/core/src/value/string.rs
+++ b/core/src/value/string.rs
@@ -6,14 +6,28 @@ use std::{mem, slice, str};
 #[repr(transparent)]
 pub struct String<'js>(pub(crate) Value<'js>);
 
+impl<'js> AsRef<str> for String<'js>{
+    fn as_ref(&self) -> &str {
+        self.as_str().unwrap()
+    }
+}
+
 impl<'js> String<'js> {
     /// Convert the JavaScript string to a Rust string.
     pub fn to_string(&self) -> Result<StdString> {
-        self.as_str().map(|s| s.into())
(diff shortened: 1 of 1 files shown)


Sytten (commented):
I would also implement `AsRef<str>`, its always a nice thing to have

richarddd:
> I would also implement `AsRef<str>`, its always a nice thing to have

I'm not to happy about the required unwrap(). I leave this as an exercise for the users instead.

DelSkayn:
This happens to work in the tests but the code actually is a use after free. ToCStringLen allocates a new buffer which is then later freed by qjs::JS_FreeCString invalidating the `&str` returned but `as_str`.

I am guessing it works in the test because the memory is not yet overwritten but valgrind correctly flags it as a use after free.

Sytten:
Wow how did I missed that. This is unsound all the way. I checked the C code to see what it does. In case of an ASCII string it wont allocate but it will if the string is an utf-8 or utf-16. There is no way around it.

richarddd:
> This happens to work in the tests but the code actually is a use after free. ToCStringLen allocates a new buffer which is then later freed by qjs::JS_FreeCString invalidating the `&str` returned but `as_str`.
> 
> I am guessing it works in the test because the memory is not yet overwritten but valgrind correctly flags it as a use after free.
> 
> Here is a short snippet showing this is unsound.
> 

I've modified to only free c string after converting to string but looking at the c source now this will leak memory. I don't think there is a way to return a string slice since we allocate in c

PR #372, opened 2024-10-08
Add CString implementation

Sytten:
This allows users to avoid a double allocation for strings when you don't need to mutate the string.

### DIFF
--- a/core/src/value/string.rs
+++ b/core/src/value/string.rs
@@ -1,5 +1,5 @@
 use crate::{qjs, Ctx, Error, Result, StdString, Value};
-use std::{mem, slice, str};
+use std::{ffi::c_char, mem, ptr::NonNull, slice, str};
 
 /// Rust representation of a JavaScript string.
 #[derive(Debug, Clone, PartialEq, Hash)]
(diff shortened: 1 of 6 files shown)


core/src/value/string.rs
```diff
@@ -37,6 +42,66 @@ impl<'js> String<'js> {
     }
 }
 
+/// Rust representation of a JavaScript C string.
+#[derive(Debug)]
+pub struct CString<'js> {
+    ptr: *const c_char,
```
DelSkayn, line 48: This should probably be a `NonNull`.
Feels more correct as this pointer should never be null and might allow for more optimizations within enums.

core/src/value/string.rs
```diff
@@ -37,6 +42,66 @@ impl<'js> String<'js> {
...
+    len: usize,
+    ctx: Ctx<'js>,
+}
+
+#[allow(dead_code)]
```
DelSkayn, line 53: Why is this attribute here?

DelSkayn (commented):
Looks good overall! Just some minor comments.

core/src/value/string.rs
Sytten: Agreed

core/src/value/string.rs
Sytten: Bad copy paste

richarddd:
@Sytten any updates on this? Would be super useful in LLRT where we just need a reference without need to mutate or copy.

Sytten:
I can push an updated version today

Sytten:
@DelSkayn @richarddd Ready for review

DelSkayn (approved):

Issue #340, opened 2024-07-26
ClassId clash among instances of generic types


FrankHB:
Expansion of `#[class]` will have a `static` objects `fn class_id() `. Unfortunately, there is still one instance when the struct is generic, due to the fact Rust share the static object for all instances (which is different in some other languages like C++). This would make very nasty bugs when more than one instances are registered to one runtime instance, when the wrong finalizers are eventually called unexpectedly.

(Also note the generated instances of `NAME` should not clash.)

A workaround is to reimplement the result from the `class` macro with injected members from additional traits describing the `NAME` and non-shared `fn class_id` implementations. This still meets the difficulties when used across different crates for problems raised by orphan rules. Thus, I believe it is better to be fixed in the crate `rquickjs_macro`.

DelSkayn:
Hmmm, this definitely a problem. I wasn't aware of this behavior.

I don't see a good way around this problem, the best solution I see is just disallowing generics on the class attribute, and force people to implement the class method on fully specified types, maybe allowing implementing it on type aliases.

Could you maybe expand on the workaround you mentioned? 

The instances of name should clash I think. The implementation of class for a generic struct is
```rust
struct Foo<T>{ t: T};
impl<'js,T> JsClass<'js> for Foo<T>{
    const NAME: &'static str = "Foo";
}
```
So name is the same for every T.
 I already expected them to class but didn't think it was that much of a problem as it the NAME wasn't critical for safety.

FrankHB:
# Current workaround

In my case I need some types to get `Rc<RefCell<T>>` work (passed around JS and Rust code) with many `T`. I have a wrapper struct and a templated mixin like this:


```rust
#[derive(derivative::Derivative)]
#[derivative(Clone(bound = ""))]
pub struct W<T: ?Sized> {
    pub value: Rc<RefCell<T>>
}
// XXX: #[rquickjs::class] does not work here. It seems also clashing with tuple struct.
// ... impl as #[rquikjs::class] omitted.

pub trait ClassDesc
{
	const NAME: &'static str;

	fn class_id() -> &'static ClassId;
}

macro_rules! define_desc {
	($ty: ty) => {
		impl $crate::ClassDesc for $ty {
			const NAME: &'static str = stringify!($ty);

			fn class_id() -> &'static $crate::rquickjs::class::ClassId {
				static ID: $crate::rquickjs::class::ClassId
					= $crate::rquickjs::class::ClassId::new();

				&ID
			}
		}
	}
[...]

PR #373, opened 2024-10-10
Rework the class system

DelSkayn:
This PR changes how the classes work in rquickjs. 

Currently rquickjs uses the internal QuickJS class registration system to handle rust classes.
This requires some way to store class-id's per type. 
In the existing code this is done with a per type static global which is initialized once and then reused.
This works because QuickJS class ids are global and don't depend on the runtime. 

However this approach has some problems.
First in QuickJS-NG the class-ids are no longer global which breaks this approach.
Second the class-ids globals are not actually unique per type but per implementation of the `JsClass` trait. 
This causes problems for generic types like in #340.

This PR instead uses only two QuickJS classes registered during initialization: rust-class and callable-rust-class. 
These classes are defined to use a v-table stored alongside allocated rust class objects to handle finalization, tracing and calling in the case of callable rust classes.
This v-table can also be used to distinguish between different classes and different types and works for generic objects fixing #340.
Prototypes are handled fully on the rust side, whenever a class is instantiated its prototype is looked up in a hashmap stored within the runtime.

### DIFF
--- a/core/src/allocator.rs
+++ b/core/src/allocator.rs
@@ -7,10 +7,6 @@ mod rust;
 
 pub use rust::RustAllocator;
 
-/// Raw memory pointer
-#[cfg_attr(feature = "doc-cfg", doc(cfg(feature = "allocator")))]
-pub type RawMemPtr = *mut u8;
-
 /// The allocator interface
 ///
 /// # Safety
@@ -26,28 +22,28 @@ pub unsafe trait Allocator {
     /// Allocate new memory
     ///
     ///
-    fn alloc(&mut self, size: usize) -> RawMemPtr;
+    fn alloc(&mut self, size: usize) -> *mut u8;
 
     /// De-allocate previously allocated memory
     ///
     /// # Safety
     /// Caller must ensure that the pointer that is being deallocated was allocated by the same
     /// Allocator instance.
-    unsafe fn dealloc(&mut self, ptr: RawMemPtr);
+    unsafe fn dealloc(&mut self, ptr: *mut u8);
 
     /// Re-allocate previously allocated memory
     ///
     /// # Safety
     /// Caller must ensure that the pointer points to an allocation that was allocated by the same
     /// Allocator instance.
-    unsafe fn realloc(&mut self, ptr: RawMemPtr, new_size: usize) -> RawMemPtr;
+    unsafe fn realloc(&mut self, ptr: *mut u8, new_size: usize) -> *mut u8;
 
     /// Get usable size of allocated memory region
     ///
     /// # Safety
     /// Caller must ensure that the pointer handed to this function points to an allocation
     /// allocated by the same allocator instance.
-    unsafe fn usable_size(ptr: RawMemPtr) -> usize
+    unsafe fn usable_size(ptr: *mut u8) -> usize
     where
         Self: Sized;
 }
@@ -140,7 +136,7 @@ impl AllocatorHolder {
         let state = &mut *state;
         state.malloc_count -= 1;
 
-        let size = A::usable_size(ptr as RawMemPtr);
+        let size = A::usable_size(ptr as *mut u8);
 
         let allocator = &mut *(state.opaque as *mut DynAllocator);
         allocator.dealloc(ptr as _);
@@ -167,7 +163,7 @@ impl AllocatorHolder {
             return ptr::null_mut();
         }
 
-        let old_size = Self::size_t(A::usable_size(ptr as RawMemPtr));
+        let old_size = Self::size_t(A::usable_size(ptr as *mut u8));
 
         let new_malloc_size = state_ref.malloc_size - old_size + size;
         if new_malloc_size > state_ref.malloc_limit {
@@ -181,7 +177,7 @@ impl AllocatorHolder {
             return ptr::null_mut();
         }
 
-        let actual_size = Self::size_t(A::usable_size(ptr as RawMemPtr));
+        let actual_size = Self::size_t(A::usable_size(ptr as *mut u8));
 
         state_ref.malloc_size -= old_size;
         state_ref.malloc_size += actual_size;
--- a/core/src/allocator/rust.rs
+++ b/core/src/allocator/rust.rs
@@ -3,7 +3,7 @@ use std::{
     mem, ptr,
 };
 
-use super::{Allocator, RawMemPtr};
+use super::Allocator;
 
 /// The largest value QuickJS will allocate is a u64;
 /// So all allocated memory must have the same alignment is this largest size.
@@ -35,7 +35,7 @@ fn round_size(size: usize) -> usize {
 pub struct RustAllocator;
 
 unsafe impl Allocator for RustAllocator {
-    fn alloc(&mut self, size: usize) -> RawMemPtr {
+    fn alloc(&mut self, size: usize) -> *mut u8 {
         let size = round_size(size);
         let alloc_size = size + HEADER_SIZE;
 
@@ -57,7 +57,7 @@ unsafe impl Allocator for RustAllocator {
         }
     }
 
-    unsafe fn dealloc(&mut self, ptr: RawMemPtr) {
+    unsafe fn dealloc(&mut self, ptr: *mut u8) {
         let ptr = ptr.sub(HEADER_SIZE);
         let alloc_size = ptr.cast::<Header>().read().size + HEADER_SIZE;
         let layout = Layout::from_size_align_unchecked(alloc_size, ALLOC_ALIGN);
@@ -65,7 +65,7 @@ unsafe impl Allocator for RustAllocator {
         alloc::dealloc(ptr, layout);
     }
 
-    unsafe fn realloc(&mut self, ptr: RawMemPtr, new_size: usize) -> RawMemPtr {
+    unsafe fn realloc(&mut self, ptr: *mut u8, new_size: usize) -> *mut u8 {
         let new_size = round_size(new_size);
 
         let ptr = ptr.sub(HEADER_SIZE);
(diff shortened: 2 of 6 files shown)


richarddd:
I really like this design - it feels much simpler and less object-oriented. While we're revisiting the class and function FFI, have you considered using polymorphism through Rust or JavaScript classes? I know it's out of scope for this PR, but I wanted to mention it so we don't make future integration more complex. I would add some test cases for extending Rust classes in JS (works in current class implementation):
```javascript
class Vec4 extends Vec3 {
   w = 0
   constructor(x,y,z,w){
      super(x,y,z);
      this.w = w;
   }
}
```

richarddd:
This is super nice. Once this is merged I will rebase my NG PR and we should be pretty good to go 👌

DelSkayn:
Unfortunately it seems I made an invalid assumption which makes this PR as it is currently unsound.

It turns out that constant address are not guaranteed to be stable nor unique. So the type distinction by v-table address is unsound. See rust-lang/unsafe-code-guidelines#522. This is probably the reason why the windows test fails. 

This could still work if there was a way to get store a type-id within the v-table. However TypeId::of::<T>() is not yet const and it requires T: 'static which is not compatible with the 'js lifetime safety mechanism. So we would need an extra function stored in the v-table which returns the type-id for the 'js lifetime erased type. 

I am currently experimenting with adding a trait called `ClassId` which is unsafe to implement and has a single function `id` which returns the TypeId for the `'js` erased type.

richarddd:
> This could still work if there was a way to get store a type-id within the v-table. However `TypeId::of::<T>()` is not yet const and it requires `T: 'static` which is not compatible with the `'js` lifetime safety mechanism. So we would need an extra function stored in the v-table which returns the type-id for the `'js` lifetime erased type.

What about using the approach i did to store an AtomicUsize in a TypeId struct with a const new function and initialize the counter via fetch_update? That way we get a thread safe alternative to `TypeId::of::<T>()` that has a static reference:
https://github.com/DelSkayn/rquickjs/blob/6cf07619b1a1c159933ba7f75d440bc97e842a25/core/src/class/id.rs#L8-L10

https://github.com/DelSkayn/rquickjs/blob/6cf07619b1a1c159933ba7f75d440bc97e842a25/core/src/class/id.rs#L35-L43

richarddd:
> > What about using the approach i did to store an AtomicUsize in a TypeId struct with a const new function and initialize the counter via fetch_update? That way we get a thread safe alternative to TypeId::of::() that has a static reference:
> 
> Does this still use the old approach of having JsClass have a function `fn class_id() -> &'static ClassId` which returns the static reference? Because this approach is broken. It doesn't return a static reference per type but per implementation so if you do.
> 
> ```rust
> struct Generic<T>(T);
> 
> impl<'js,T> JsClass<'js> for Generic<T> {
>     const NAME: &'static str = "Generic";
>     type Mutable = Readable;
>     const CALLABLE: bool = true;
>     
>     fn class_id() -> &'static crate::class::ClassId {
>         static ID: ClassId = ClassId::new();
>         &ID
>     }
> 
>     fn prototype(ctx: &Ctx<'js>) -> Result<Option<Object<'js>>> {
>        // .. omit other functions.
>     }
> }
> ```
> 
> Then
> 
> ```rust
> std::ptr::eq(<Generic<usize> as JsClass>::class_id(), <Generic<u8> as JsClass>::class_id())
> ```
> 
> Returns true because all `Generic<T>` share the same implementation. So we can't relay on this to ensure two types are the same.

Ah right, i didn't consider generics. Can we `transmute(TypeId::of::<T>())` to get rid of static?

```rust
pub struct TypeId {
    pub val: (u64, u64),
}
impl TypeId {
    pub fn of<T: Any>() -> Self {
        unsafe { mem::transmute(TypeId::of::<T>()) }
    }
}

```

DelSkayn:
> Ah right, i didn't consider generics. Can we transmute(TypeId::of::<T>()) to get rid of static?

The static problem is in that to get the type id of `T` `T` needs to be `'static` as `Any` has `: 'static` as a trait bound. So if you have a type which contains an rquickjs object like `Container<'js>(Object<'js>)` you need that `'js` lifetime on the container type which in turn means it isn't `'static` which means we can't get the type id of that type. 

So you wouldn't be able to obtain a `Class<'js,Container<'js>>` as we can't have none static types if we directly rely on `Type::of::<T>()` to get the type id.

DelSkayn:
I have an approach where I require all implementers of `JsClass` to also implement `Outlive` which now has the definition of:
```rust
pub unsafe trait Outlive<'js> {
    /// The target which has the same type as a `Self` but with another lifetime `'t`
    type Target<'to>: 'to;
}
```
Outlive was previously used to erase the lifetime of types in Persistant so that the could be handled outside of the `with` closure. The only change is the extra `'to` bound which any type which implemented this trait should fit. If all classes implement this trait you can just do `TypeId::of::<Cls::Target<'static>>()` to get the type id.

core/src/runtime/opaque.rs
```diff
@@ -37,24 +51,63 @@ impl<'js> Opaque<'js> {
...
+            call: None,
+            exotic: ptr::null_mut(),
+        };
+
+        if 0 != qjs::JS_NewClass(rt, self.class_id, &class_def) {
+            return Err(Error::Unknown);
         }
+
+        let class_def = qjs::JSClassDef {
+            class_name: b"RustFunction\0".as_ptr().cast(),
```
richarddd, line 95: `RustFunction`  => `__qjs_rf`

core/src/runtime/opaque.rs
```diff
@@ -37,24 +51,63 @@ impl<'js> Opaque<'js> {
...
+        this.spawner = Some(UnsafeCell::new(Spawner::new()));
+        this
+    }
+
+    pub unsafe fn initialize(&mut self, rt: *mut qjs::JSRuntime) -> Result<(), Error> {
+        qjs::JS_NewClassID((&mut self.class_id) as *mut qjs::JSClassID);
+        qjs::JS_NewClassID((&mut self.callable_class_id) as *mut qjs::JSClassID);
+
+        let class_def = qjs::JSClassDef {
+            class_name: b"RustClass\0".as_ptr().cast(),
```
richarddd, line 83: Should we also use a more private name, i.e `__qjs_rc` and put this as a const some where?

core/src/runtime/opaque.rs
```diff
@@ -25,6 +32,13 @@ pub(crate) struct Opaque<'js> {
     /// The user provided interrupt handler, if any.
     interrupt_handler: UnsafeCell<Option<InterruptHandler>>,
 
+    /// The class id for rust classes.
+    class_id: qjs::JSClassID,
+    /// The class id for rust classes which can be called.
+    callable_class_id: qjs::JSClassID,
+
+    prototypes: UnsafeCell<HashMap<*const VTable, Option<Object<'js>>>>,
```
richarddd, line 40: As we'll always add prototypes and never remove them we can use TinyVec instead and store the index in the VTable. Something  like this (psudocode):
```rust
prototypes: Unsafecell<TinyVec<[(*const VTable, Option<Object<'js>>); 1024]>>
```
```rust
pub fn get_or_insert_prototype<C: JsClass<'js>>(
        &self,
        ctx: &Ctx<'js>,
    ) -> Result<Option<Object<'js>>, Error> {
        unsafe {
            let vtable = VTable::get::<C>();
            let prototypes =  (*self.prototypes.get());
            if let Some(index) = vtable.index {
                return prototypes[index].1
            }
            let next_index = prototypes.length();
            vtable.index = Some(next_index);
            prototypes.push((vtable,C::prototype(ctx)?)
        }
    }
```

richarddd (changes requested):
Some minor comments

core/src/runtime/opaque.rs
richarddd: Just occurred to me that this might also be a bit more complex and not feasible as we use the same vtable for many runtimes.

richarddd: So think i figured out a great way to solve this without hashtable:

1. Each vtable gets an additional auto incrementing index_id (derived form an atomic u32)
2. `prototypes` are stored as a `Vec` (or `TinyVec`)
3. Proto objects are stored and acceced via the array index of index_id for the vtable

Fast, constant time lookup, no hashing or collitions

richarddd: We can use trait top level recursion to create unique auto incrementing IDs at compile time. I think this could work and has zero cost
```rust
pub trait Counter {
    const NEXT: usize;
}

impl Counter for () {
    const NEXT: usize = 0;
}

//we don't care about T just want to increment the value
pub struct Increment<T>(std::marker::PhantomData<T>);

impl<T: Counter> Counter for Increment<T> {
    const NEXT: usize = T::NEXT + 1;
}
```

```rust
const VTABLE: VTable = {
  // generate unique ID for this VTable using type-level counter
  const ID: usize = <Increment<Increment<()>> as Counter>::NEXT;
  
  VTable {
      id: TypeId::of::<C::Changed<'static>>,
      vtable_id: ID,  // Assign unique compile-time ID
      finalizer: VTable::finalizer_impl::<'js, C>,
      trace: VTable::trace_impl::<C>,
      call: VTable::call_impl::<C>,
  }
};
```

core/src/runtime/opaque.rs
DelSkayn: We can, but it doesn't really matter. `RustClass` is not accessable anywhere, it is not stored in the global object or anywhere else so it can't pollute a namespace. 
It also can't cause conflicts as you can register two classes with the same class_name without problems. 

The only place you will see this name is in dumps when enabling the debug dump features. It might be a good idea to make this a bit more descriptive, maybe `RquickjsClass` to clearly signify where the class came from.

PR #382, opened 2024-11-06
Only increment userdata count if we actually have it

richarddd:
If we don't have specific userdata, we still increment the count causing errors if we then try to insert it.


DelSkayn (approved):
Good catch!

PR #385, opened 2024-11-12
Fix/prototype lookup causing prototype missmatch across crates

richarddd:
When running tests for LLRT i discovered that classes living in different crates got different TypeId function pointers. This messed up the prototypes. Invoking the TypeId function pointer however resulted in the same ID. This may have to do with optimizations and rust compiler internals.

Safest thing to do with now be to use the actual TypeId as key for the prototype table.


DelSkayn (approved):
Good catch, missed this when I encountered the same problem with matching classes. Must have forgotten to change the code for prototypes.

PR #393, opened 2024-11-19
fix: realloc rust allocator memory leak

richarddd:
When callculating `alloc_sice` for realloc we fail to account for the header size. This results in memory errors & leaks.
This PR fixes those issues.

### DIFF
--- a/core/src/allocator/rust.rs
+++ b/core/src/allocator/rust.rs
@@ -41,10 +41,8 @@ unsafe impl Allocator for RustAllocator {
         }
 
         let total_size = count.checked_mul(size).expect("overflow");
-
         let total_size = round_size(total_size);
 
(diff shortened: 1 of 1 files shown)


DelSkayn (approved):
Yes good catch!

Sytten:
@DelSkayn I was suggesting to @richarddd that we should try to run miri on rquickjs in CI to catch bugs in unsafe code. Thoughts?

DelSkayn:
Miri does not work with C FFI. It is mostly for pure rust code, the moment you call out to a C function miri does not know, i.e. not a part of libc, it gives up.

DelSkayn:
We can look into running ASAN in CI, that might work but I am not sure how to get it working on both the rust and C side.=

PR #399, opened 2024-11-27
Impl deref_mut for UserdataGuard

richarddd:
Should be safe since we're locking runtime when accessing context


richarddd:
Probably needs a bit more work, transmute to &mut is UB

PR #407, opened 2024-12-24
Use new APIs from QJS to create and lookup typed arrays

richarddd:
This allows us to construct and lookup typed arrays by an int type which does direct comparison to the hardcoded class_id value stored in QJS. It's also safer since the global value can be overwritten in js user space


Sytten:
We are likely missing the bindings for other platforms

richarddd:
Yeah, CI is a bit broken as it requires bindings to run the checks. I will manually add them, but it's time consuming.

richarddd:
@DelSkayn please LMKWYT regarding this one

DelSkayn (approved):
Looks good

PR #415, opened 2025-01-21
Fix is_typed_array

richarddd:
The current implementation panics if object is not a typed array due to bad unwrap. This PR fixes that.
Also fixes https://github.com/DelSkayn/rquickjs/issues/416


richarddd:
@DelSkayn please have a look. I had to bump MSRV to 1.77 because some features requires minimum 1.77. Also bindings generate `bool` on some platforms and `i32` on others, hence the required casting.

DelSkayn:
Weird that the MSRV didn't fail before. If offset_of was introduced in 1.77 it shouldn't have compiled on 1.65..

.gitignore
```diff
@@ -1,4 +1,3 @@
 target
 .ccls-cache
-Cargo.lock
```
DelSkayn, line 3: I would like to avoid committing the lock file. I think it is still best to ensure rquickjs is tested against the latest dependencies if possible. The linked blog post says that they do not have a default recommendation not that they recommend against not committing the lockfile.

richarddd:
> Weird that the MSRV didn't fail before. If offset_of was introduced in 1.77 it shouldn't have compiled on 1.65..

I didn't fail because that code was feature gated

> I would like to avoid committing the lock file. I think it is still best to ensure rquickjs is tested against the latest dependencies if possible. The linked blog post says that they do not have a default recommendation not that they recommend against not committing the lockfile.

No problem, but then MSRV needs to be bumped to 1.81 due to `home` dependency:
https://crates.io/crates/home/versions
https://github.com/DelSkayn/rquickjs/actions/runs/12889992657/job/35938535528

DelSkayn (approved):
Looks good!

PR #422, opened 2025-01-27
expose promise rejection tracker

Gentle:
this enables users to write a custom onerror handler for uncaught rejected promises

the parameters are currently the same as the C API has, this function receives both handled and unhandled rejections.

For the purpose of onerror, you would ignore all calls where is_handled is true as below

example usage:
```rust
use rquickjs::{Context, Runtime};

fn main() {
    let runtime = Runtime::new().unwrap();
    let context = Context::full(&runtime).unwrap();
    runtime.set_host_promise_rejection_tracker(Some(Box::new(
        |_ctx, _promise, reason, is_handled| {
            if !is_handled {
                dbg!(reason);
            }
        },
    )));
    context
        .with(|ctx| {
            let _: () = ctx
                .eval("function x() { new Promise(() => { throw new Error('42') }) }; x()")
                .unwrap();
        });
}

```

### DIFF
--- a/core/src/runtime/raw.rs
+++ b/core/src/runtime/raw.rs
@@ -13,10 +13,10 @@ use crate::allocator::{Allocator, AllocatorHolder};
 use crate::loader::{Loader, LoaderHolder, Resolver};
 use crate::{
     qjs::{self, size_t},
-    Error, Result,
+    Ctx, Error, Result, Value,
 };
 
-use super::{opaque::Opaque, InterruptHandler};
+use super::{opaque::Opaque, InterruptHandler, RejectionTracker};
 
 const DUMP_BYTECODE_FINAL: u64 = 0x01;
 const DUMP_BYTECODE_PASS2: u64 = 0x02;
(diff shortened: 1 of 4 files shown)


Gentle:
where would I put a test for this?

Gentle:
ok, while this works, it causes ref_count errors, I got these logs when debugging

```
Assertion failed: p->ref_count == 0 (/Users/gentle/work/quantum/quantum-native/target/wasm32-wasip1/wasm-release/build/rquickjs-sys-c03be711d9001ed4/out/quickjs.c
free_zero_refcount: 5617)
```

I'll try debugging further tomorrow, do you have any insights on what needs to be changed?

Gentle:
the current iteration is sometimes causing the garbage collector to crash when shutting down, but otherwise does the job

Gentle:
I'm happy to say I fixed it, I'll write a test tomorrow

core/src/runtime/raw.rs
```diff
@@ -331,3 +366,38 @@ impl RawRuntime {
...
+                const x = async () => {
+                    throw new Error("Uncaught")
+                }
+                x()
+                throw new Error("Caught")
+            "#,
+            );
+        });
+        assert_eq!(*counter.borrow(), 1);
+    }
```
DelSkayn, line 402: This test breaks when the parallel feature is enabled, requiring send the rejection closure. If you change the `Rc` to `Arc<AtomicUsize>` it will fix it.

DelSkayn (commented):
Looks good, just need a change to the test to make sure it runs when `parallel` is enabled.

core/src/runtime/raw.rs
Gentle: fixed in next commit :)

DelSkayn (approved):
LGTM

PR #429, opened 2025-02-15
Fix namespace resolution for JsLifetime in derive macro

Digifox03:
The derive macro for JsLifetime does not always use the fully qualified name.


DelSkayn (approved):
Looks good, Thanks for the PR!

PR #441, opened 2025-03-08
`wasm32-wasip1`/`wasm32-wasip2` tests

lastmjs:
This PR enables as many of the automated CI test suite as possible for the `wasm32-wasip1` and `wasm32-wasip2` targets.


lastmjs:
I'm opening up this draft to get some feedback. Most of the test suite from `.github/workflows/ci.yml` is passing. But there are 13 tests from `Doc-tests rquickjs_core` that aren't running. I'd like some guidance on if these `Doc-tests` would be valuable to run, and if we should figure out some way to get this to run. My feeling after my current round of R&D for this would be that `Doc-tests` aren't really possible to run on `wasm32` targets, and that it would be more practical to extract those tests and run them normally, if the test coverage for those tests is desirable.

DelSkayn:
I personally see doc tests as testing for the validity of examples given in the docs, ensuring the code there actually runs. 
Useful to test for but not really necessary to run on every platform.

DelSkayn (approved):
Looks good! Thanks for the work!

lastmjs:
Awesome!

PR #443, opened 2025-03-15
Add `Module::write` options

giuliano-macedo:
Added `WriteOptions` struct that represents all flags that can be passed to `qjs::JS_WriteObject`, this is useful specially for the flags `JS_WRITE_OBJ_STRIP_SOURCE` and `JS_WRITE_OBJ_STRIP_DEBUG` which when enabled decreases the bytecode size significantly (see https://github.com/quickjs-ng/quickjs/pull/388), while adding it I've noticed two other flags (`JS_WRITE_OBJ_REFERENCE` and `JS_WRITE_OBJ_SAB`) that can be passed to it, so I've added as well ([ref](https://github.com/quickjs-ng/quickjs/blob/v0.8.0/quickjs.h#L875-L880)).


DelSkayn (approved):
Looks good, thanks for the PR!

PR #447, opened 2025-04-03
feat: Add lossy conversion between QJS string and Rust string

richarddd:
`"🌍🌎🌏".slice(1)` currently produces a broken surrogate pair. When converting to a rust string, we get UTF8 error.
QJS handles this by just printing the invalid utf16 surrogate pair as a replacement char.

This PR introduces a new method `to_string_lossy` that will render a utf8 replacement char for broken surrogate pairs.


Sytten:
@richarddd I would argue that we should probably start using the new UTF-16 API instead https://github.com/quickjs-ng/quickjs/issues/992 and do all the conversion on the rust side.

Sytten:
We now have from and to utf16 from the engine, we should use that

PR #452, opened 2025-04-20
Add missing set_host_promise_rejection_tracker to AsyncRuntime

Jmgr:
This adds set_host_promise_rejection_tracker to the AsyncRuntime.


nabetti1720:
The promise hook API created in #453 doesn't allow us to detect rejected promises, but can we use this to do that?

richarddd (approved):

PR #455, opened 2025-04-23
Allow rquickjs-core to build with no_std

### LINKED ISSUE #303: Support no_std targets
It'd be great if I could use the `rquickjs` crate on no_std targets as an embedded JavaScript engine.

jprendes:
This PR allows rquickjs-core to build with no_std.

* The rquickjs-sys crate is made unconditionally no_std.
* The rquickjs-core crate can be made no_std by disabling the new `std` feature, which is enabled by default.
* The FileResolver in the loader module had to be disabled building with no_std.
* ~~The relative-path dependency had to be vendored in and edited to be no_std compatible. When the upstream changes at https://github.com/udoprog/relative-path/pull/64 are merged, we can go back to using the upstream dependency directly.~~ Upstream now supports no_std

Fixes #303


jprendes:
@DelSkayn PTAL :-)

jprendes:
Fixed formatting, sorry about that!

jprendes:
@DelSkayn PTAL :-)

jprendes:
@DelSkayn PTAL :-)

richarddd:
@jprendes please rebase and me and @Sytten will take a look :)

jprendes:
Rebased

It's great that you two are getting actively involved in maintaining the repo, thanks!

I also manually fixed the bindings, I don't know if CI uses the checked-in bindings or regenerates them every time.
Anyhow, with this PR the bindings generation should generate bindings for core: https://github.com/DelSkayn/rquickjs/pull/455/files#diff-2620be2afc249413fa14ee1f678c9c7ab8f0658f44f5c6d49cff57956c716deb

Sytten:
@jprendes CI will regenerate them once merged to main. But this is fine since you modified the build script it should be fine.

Sytten:
@jprendes We will need to add a test case with std disabled so we dont break it in the future, also please check the failing format.

jprendes:
> We will need to add a test case with std disabled

I think this is already the case most of the time, as CI as using `--no-default-features`, which would disable `std` (except when re-enabled by some other feature like `full-xxx`)

richarddd (approved):

PR #497, opened 2025-08-30
Add support for 'exotic objects' (exotic field in JSClassDef) to JSClass, remove Send bound from async_with

cheesycod:
Add support for 'exotic objects' (exotic field in JSClassDef) to JSClass, remove Send bound from async_with.

Exotic objects allow for classes with custom get/set/has/delete property implementations and are hence non-exhaustive in property listing. This is important for my use case of proxying tables+userdata between Luau and QuickJS. Other options/alternatives for this are to use a Proxy but this has slightly more memory usage, perf loss, some GC related concerns and type erasure (no longer as easy to map the table/userdata back from QuickJS to Luau for example as the Proxy 'erases' the real inner type) compared to the feature that already exists in quickjs itself.

**Note: This PR only implements get_property, set_property, has_property and delete_property as these are the simplest (by far) to implement and also the most important**

### DIFF
--- a/core/src/class.rs
+++ b/core/src/class.rs
@@ -54,7 +54,7 @@ pub trait JsClass<'js>: Trace<'js> + JsLifetime<'js> + Sized {
     }
 
     /// The function which will be called if a get property is performed on an object with this class
-    fn exotic_get_property<'a>(
+    fn exotic_get_property(
         this: &JsCell<'js, Self>,
         ctx: &Ctx<'js>,
         _atom: Atom<'js>,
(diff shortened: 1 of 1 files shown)


richarddd:
@cheesycod can you take a look. Tests are failing :)

cheesycod:
> @cheesycod can you take a look. Tests are failing :)

Yeah, will do. I honestly forgot about this PR I made entirely lol

richarddd:
@cheesycod i think the no-std build support is making this fail, can you review?

cheesycod:
> @cheesycod i think the no-std build support is making this fail, can you review?

yeah, will do this week (I promise!), i just got super busy with side projects etc.

richarddd:
@cheesycod please take a stab at this when you have time, converted to draft for now!

Sytten:
I might be interested in this PR for a project of ours to avoid a Proxy indirection.

Sytten:
Closing since we have a PR ready

PR #506, opened 2025-09-21
fix(build): strip llvm suffix for *-pc-windows-gnullvm target

ognevny:
otherwise bindgen fails because clang reports this suffix as wrong in target name


richarddd (approved):

ognevny:
CI failure seems unrelated to my change

PR #535, opened 2025-10-22
Add disable-assertions Cargo feature to disable C assertions

### LINKED ISSUE #528: Feature to define NDEBUG for enhanced performance
Would you be open to adding a Cargo feature to define `NDEBUG` when compiling QuickJS to disable the C assertions in QuickJS?

We use rquickjs to embed a QuickJS interpreter inside a Wasm module and these Wasm instances are short-lived and are subject to aggressive limits on the number of instructions they're allowed to execute. By disabling C assertions by defining `NDEBUG`, we'd be able to use fewer Wasm instructions to evaluate the same JavaScript code.

jeffcharles:
Closes #528. Adds a Cargo feature named `disable-assertions` that will add a `NDEBUG` definition to the QuickJS build which strips out C assertions. I've validated this uses a materially lower number of Wasm instructions to execute the same JavaScript code with the same inputs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -250,6 +250,11 @@ jobs:
             rust: stable
             target: x86_64-unknown-linux-gnu
             features: parallel
+          - task: features
+            os: ubuntu-latest
+            rust: stable
+            target: x86_64-unknown-linux-gnu
+            features: disable-assertions
 
           # Test channels
           - task: channels


Sytten (approved):

Sytten:
Ideally we should also create a check in GHA to ensure we dont break it in the future

jeffcharles:
Agreed that would be nice but I'm not sure what would be a good way to test the absence of assertions in this test suite though. The reason I'm confident this works is we have performance tests in Javy and the numbers look much lower when I use this as a local dependency and enable the Cargo feature compared to when I don't enable the feature on the same code.

richarddd:
> Agreed that would be nice but I'm not sure what would be a good way to test the absence of assertions in this test suite though. The reason I'm confident this works is we have performance tests in Javy and the numbers look much lower when I use this as a local dependency and enable the Cargo feature compared to when I don't enable the feature on the same code.

We could add an additional test run here:
https://github.com/DelSkayn/rquickjs/blob/b1bc6a3592d82303f17e1a2ef38ad8ecb1bddd6e/.github/workflows/ci.yml#L64

Add below current:
```bash
cargo clippy --all --all-targets --features full-async,bindgen,disable-assertions
```

jeffcharles:
Could you clarify what you would like in that additional test run?

Do you mean run clippy a second time with that feature enabled? Or run `cargo test` with that feature enabled? Or something else?

Sytten:
@jeffcharles Actually it would be good to run the test suite instead.
Adding a case to the matrix [here](https://github.com/DelSkayn/rquickjs/blob/master/.github/workflows/ci.yml#L136) should be easy.
Thanks!

richarddd (approved):

Issue #540, opened 2025-10-25
New release 1.10 contains a lot of build issues


lublak:
I updated to the latest version, but now i get a lot of build issues:

```
error[E0080]: evaluation of constant value failed
   --> /home/*****/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rquickjs-sys-0.10.0/src/bindings/wasm32-wasip1.rs:215:35
    |
215 |     ["Size of JSMallocFunctions"][::core::mem::size_of::<JSMallocFunctions>() - 40usize];
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `20_usize - 40_usize`, which would overflow

error[E0080]: evaluation of constant value failed
   --> /home/*****/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rquickjs-sys-0.10.0/src/bindings/wasm32-wasip1.rs:677:38
    |
677 |     ["Size of JSClassExoticMethods"][::core::mem::size_of::<JSClassExoticMethods>() - 56usize];
    |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `28_usize - 56_usize`, which would overflow

error[E0080]: evaluation of constant value failed
   --> /home/*****/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rquickjs-sys-0.10.0/src/bindings/wasm32-wasip1.rs:720:28
    |
720 |     ["Size of JSClassDef"][::core::mem::size_of::<JSClassDef>() - 40usize];
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `20_usize - 40_usize`, which would overflow

error[E0080]: evaluation of constant value failed
   --> /home/*****/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rquickjs-sys-0.10.0/src/bindings/wasm32-wasip1.rs:741:31
    |
741 |     ["Size of JSEvalOptions"][::core::mem::size_of::<JSEvalOptions>() - 24usize];
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `16_usize - 24_usize`, which would overflow

[...]

PR #548, opened 2025-10-30
build wasm32 bindgen

lublak:


lublak:
fixes for me #540 web wasm wasip1 (also tested locally)

richarddd (approved):

lublak:
@richarddd is there an eta when the wasip1 binding will be released?

PR #567, opened 2025-12-04
Exotic methods support

### LINKED ISSUE #495: Why does async_with need the future to implement Send?
Why does AsyncContext::async_with need the future inputted to be Send even when parallel feature is not used?

Sytten:
- Add support for exotic methods
- Add a macro to more easily implement those
    - it's a bit ugly but it works, I will improve it over time
    - Unsure about the get,set,has,delete names
- Remove send bound for async_with when not in parallel mode

Finishes #497


richarddd (approved):
Minor suggestions

core/src/class.rs
```diff
@@ -30,6 +31,9 @@ pub trait JsClass<'js>: Trace<'js> + JsLifetime<'js> + Sized {
     /// Is this class a function.
     const CALLABLE: bool = false;
 
+    /// Is this class exotic (e.g. will exotic_* methods be called with it)
+    const EXOTIC: bool = false;
+
```
DelSkayn: Since a class can't be both exotic and callable, it is probably better to encode that in the type system, i.e. change these bools to something like `const KIND: ClassKind = ClassKind::Plain` where `ClassKind` is something like:
```rust
enum ClassKind{
    Plain,
    Callable,
    Exotic,
}
```

DelSkayn (commented):
Two suggestions.

core/src/class.rs
Sytten: Right that is a good idea. Though I dont like the whole system of a single class ID for all rust classes.

core/src/result.rs
```diff
@@ -689,6 +700,20 @@ impl<'js> Ctx<'js> {
...
+        F: FnOnce() -> qjs::c_int + UnwindSafe,
+    {
+        match crate::util::catch_unwind(f) {
+            Ok(x) => x,
+            Err(e) => unsafe {
+                self.get_opaque().set_panic(e);
+                qjs::JS_Throw(self.as_ptr(), qjs::JS_MKVAL(qjs::JS_TAG_EXCEPTION, 0));
+                -1
+            },
+        }
```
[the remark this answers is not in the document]
Sytten: Kinda of in a crunch right now, I need to refactor the macro and the bool into an enum. If you want to help be my guess

xaptronic:
hey, came across this while optimizing Rust/JS interop patterns. will this PR include support for get_own_property and get_own_property_names? I noticed they're TODO in ExoticMethodsHolder — those are the main callbacks we'd need. happy to help with the remaining work if useful.

Sytten:
Honestly I dont have time to work on that right now, so @richarddd if you want to take it over or have someone else take it off my plate it works for me.

xaptronic:
NP - are these `get_own_property` and `get_own_property_names` sensible / possible to do as part of this? I noticed they also don't exist on the proxy objects - would this be more sensible to do in a follow up on both?

richarddd:
> NP - are these `get_own_property` and `get_own_property_names` sensible / possible to do as part of this? I noticed they also don't exist on the proxy objects - would this be more sensible to do in a follow up on both?

@xaptronic seems in scope. will add!

Sytten:
The reason I didnt do it previously is that the vtable is starting to get pretty big. I am not as fan that all classes use the same quickjs class id since it makes not nice to extend a rust class in js.

PR #568, opened 2025-12-04
Bump msrv to 1.85

Sytten:
Because of edition 2024 of dlopen


richarddd:
Should we change edition as well?

Sytten:
Probably, need to fix the dlopen issue first

richarddd (approved):

Issue #590, opened 2026-01-04
Request to include `quickjs-libc.h` in `quickjs.bind.h`


rushmi0:
Could you please add the following include in `quickjs.bind.h`:
```c
#include "quickjs/quickjs-libc.h"
```

This will allow me to use `js_std_add_helpers` and conveniently access `console.log` and `print` in QuickJS, which will greatly improve usability.

Thank you very much for considering this request!

Sytten:
Its not just the include though, its pulling the whole libc + the dependencies of it in the build for all platforms. If you want native modules like console we are building rust native ones in llrt (https://github.com/awslabs/llrt/tree/main/modules) that you can use.

We have no immediate plan to include the quickjs-ng libc, if you want it feel free to send a PR

richarddd:
Closed due to libc deps is a runtime and this is engine bindings

PR #597, opened 2026-01-10
Adds bindings for functions declared in quickjs-libc.h.

rushmi0:
The bindings allow calling functions declared in `quickjs-libc.h`.
Ref: https://github.com/DelSkayn/rquickjs/issues/590

```rs
    #[test]
    fn test_add_std_helpers() {
        use crate::{Context, Runtime};
        use rquickjs::qjs;

        ctx.with(|ctx| {
            unsafe { qjs::js_std_add_helpers(ctx.as_ptr(), 0, core::ptr::null_mut()); }
            let script_str = r#"
                print("Hello from QuickJS");
                console.log("Hello from QuickJS");
            "#;
            if let Err(e) = ctx.eval::<(), _>(script_str).catch(&ctx) {
                panic!("{}", e);
            }

            let print_type: String = ctx.eval("typeof print").unwrap();
            assert_eq!(print_type, "function");

            let console_log_type: String = ctx.eval("typeof console.log").unwrap();
            assert_eq!(console_log_type, "function");
        });
    }
```

### DIFF
--- a/sys/build.rs
+++ b/sys/build.rs
@@ -125,6 +125,9 @@ fn main() {
 
     let header_files = [
         "builtin-array-fromasync.h",
+        "builtin-iterator-zip-keyed.h",
+        "builtin-iterator-zip.h",
+        "cutils.h",
         "dtoa.h",
         "libregexp-opcode.h",
         "libregexp.h",


Sytten:
That should probably be behind a feature flag and most likely disabled by default. What do you think @richarddd

richarddd:
I think this is strange, we cant just use quickjs-libc because thats the JS runtime and rquickjs uses it's own runtime. We have to distinguish between runtime and engine.

Sytten:
True, maybe it should be another crate? Can we inject the libc as a lib?

richarddd:
> True, maybe it should be another crate? Can we inject the libc as a lib?

I dont think so because it has its own entry point, eventloop, pool/select etc etc. It's not ment to be used outside a C runtime which rust does not include

rushmi0:
So instead of using `js_std_add_helpers`, should I just implement `print` and `console.log` directly in Rust and register them to the context?

Sytten:
You can use the llrt modules we already built most of the modules

PR #601, opened 2026-01-15
Add import attributes support to Loader trait

nitsky:
This PR adds support for import attributes (e.g., `import data from "./data.json" with { type: "json" }`) in rquickjs by updating the bindings and adding an `attributes` argument to the loader trait. This is a breaking change.

### Issue # (if available)

None

### Description of changes

- Added `JS_SetModuleLoaderFunc2`, `JSModuleLoaderFunc2`, and `JSModuleCheckSupportedImportAttributes` to FFI bindings for all 14 platforms
- Updated the `Loader` trait to add  an `attributes` argument
- Added `ImportAttributes<'js>` type alias for `Option<Object<'js>>`
- Updated `LoaderHolder` to use `JS_SetModuleLoaderFunc2`
- Updated all existing loader implementations

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Breaking Changes
+
+- Added import attributes to the Loader trait #[601](https://github.com/DelSkayn/rquickjs/pull/601)
+
 ## [0.11.0] - 2025-12-16
 
 ### Added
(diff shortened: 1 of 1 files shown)


nitsky:
@Sytten I added an example, some tests, and I made `ImportAttributes` a newtype around `Object`, with `get(&self, key: &str) -> Result<Option<String>>` and `get_type` as a convenience.

Sytten:
@nitsky is the test failure related to your changes?

nitsky:
@Sytten I don't think so, I pushed a dummy commit to try CI again, but it requires approval.

nitsky:
@Sytten more failures, but I'm not convinced these are related to this change. Any ideas how to move forward?

Sytten:
Longaarch often fails, but its been a while since stable Linux failed so I am concerned this is somehow related to this PR. Other concurrent PR dont fail on stable.

nitsky:
@Sytten It appears the most recent run succeeded. Is there anything else you would like to see changed before merge?

Sytten:
@nitsky I am updating the bindings in a separate PR #603 , you will need to rebase once its done
I would like to see a changelog line for that as a breaking change.

nitsky:
@Sytten I rebased after #603 and added a breaking change line to the changelog.

PR #602, opened 2026-01-18
Deprecate async_with! macro in favor of async closures

DelSkayn:
### Issue # (if available)

N/A

### Description of changes

The `async_with!` macro was created to work around the lack of async closures.
Without async closures it was not possible to specify the right lifetime semantics for the `async_with` function making it very limited in it's use.
Now that async closures are stable the `async_with!` macro is obsolete and should be deprecated.

This pr deprecates the `async_with!` macro and updates `AsyncContext::async_with` to use the new async closures.

### Checklist

### DIFF
--- a/tests/async_compile_fail/async_nested_contexts.stderr
+++ b/tests/async_compile_fail/async_nested_contexts.stderr
@@ -13,6 +13,10 @@ error: lifetime may not live long enough
 13 | |                     ctx_1.globals().set("t", ctx_2.globals());
 14 | |                 })
    | |_________________^ returning this value requires that `'1` must outlive `'2`
+   |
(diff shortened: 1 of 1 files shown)


Sytten:
Wasnt there a trick to lift lifetimes so that it would allow local variables to be used?

DelSkayn:
What do you mean with local variables? This does allow capturing local variables in the closure.
The following compiles fine:
```rust
    let mut var = 1u32;
    ctx.async_with(async |_ctx| {
        var += 1;
    })
    .await;
    println!("{}", var);

```

Sytten:
Ha perfect then! Yeah this now works since the compiler doesn't have to capture the variable in the callback.

PR #611, opened 2026-01-29
Add JS_UpdateStackTop calls before QuickJS C entry points

pierrekin:
### Description of changes

Fixes cross-thread stack overflow false positives by ensuring stack_top baseline is updated whenever execution moves between tokio threads during async operations. Only active when 'parallel' feature is enabled.

### Checklist


pierrekin:
I checked "I created unit tests for my change if needed" because I concluded that unit tests were NOT needed.

PR #612, opened 2026-01-29
Add JS_UpdateStackTop calls before QuickJS C entry points

pierrekin:
### Description of changes

Fixes cross-thread stack overflow false positives by ensuring stack_top baseline is updated whenever execution moves between tokio threads during async operations. Only active when 'parallel' feature is enabled.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Deprecated `async_with!` macro #[602](https://github.com/DelSkayn/rquickjs/pull/602)
 
+### Fixed
+
+- Fixed cross-thread stack overflow false positives in parallel mode by updating stack baseline before QuickJS C entry points
+
 ## [0.11.0] - 2025-12-16
 
 ### Added
--- a/core/src/context/ctx.rs
+++ b/core/src/context/ctx.rs
@@ -147,6 +147,10 @@ impl<'js> Ctx<'js> {
         let src = source.into();
         let len = src.len();
         let src = CString::new(src)?;
+
+        #[cfg(feature = "parallel")]
+        qjs::JS_UpdateStackTop(qjs::JS_GetRuntime(self.ctx.as_ptr()));
+
         let val = qjs::JS_Eval(
             self.ctx.as_ptr(),
             src.as_ptr(),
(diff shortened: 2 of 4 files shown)


pierrekin:
I checked "Created unit tests for my feature if needed" because in my judgment this does not need any _new_ unit tests.

UrsDeSwardt:
Great contribution, thanks @pierrekin! 🎉

Sytten:
Do you have a reference that this is needed from the quickjs folks? If so what are the cases where this is needed.

@richarddd have you stack overflow in multi-thread?

This seems like a big change if we suddenly need to call this every time before we call into the engine. To a point where we should reconsider if we should move the event loop into its own thread instead.

@DelSkayn also feel free to chime in

pierrekin:
Hello @Sytten, thanks for having a look at this.

> Do you have a reference that this is needed from the quickjs folks? If so what are the cases where this is needed.

No, the history here is:

1. We started getting unexpected "stack overflow" errors in our Tauri and Tokio based project.
2. I locally forked the quickjs-ng code and added some tracing and logging to dig into this.
3. I noticed that when the stack overflow errors were being raised, quickjs was internally comparing the stack pointer from one thread, against the base address from another thread.
4. I traced the execution flow through the quickjs c code, and noticed that the 'base stack pointer' to compare against is set only once unless the update stack top function is called.
5. I made the changes in this PR, and confirmed that it resolved the issue we were seeing.

I personally think that a neater fix might be to make the quickjs c code itself thread aware in the sense that it uses "thread local storage" for the base stack pointer, thus removing the need to call this 'update stack pointer base' function completely from the rust code side.

To be clear, I'm not involved in the quickjs project, or in contact with any developers from that project.

Sytten:
@richarddd I am really not sure about this PR. I must not be understanding something cause if you need to call that every time it jumps thread then in theory you would need to call it before every call to the C engine anywhere.

I am really confused why this was also never surfaced before by anybody and I would have liked to see a failing test before merging to ensure we don't remove those calls by accident later...

pierrekin:
> @richarddd I am really not sure about this PR. I must not be understanding something cause if you need to call that every time it jumps thread then in theory you would need to call it before every call to the C engine anywhere.
> 
> I am really confused why this was also never surfaced before by anybody and I would have liked to see a failing test before merging to ensure we don't remove those calls by accident later...

Totally understood @Sytten! I'll follow up with some better evidence, ideally a failing test.

Thank you for merging my PR regardless, I am sorry for taking so long to respond, I will try and prioritize this.

PR #621, opened 2026-02-11
Allow accessing the module from Declarations and Exports

jprendes:
### Description of changes

This PR exposes the underlying `Module` from the `Declarations` and the `Exports` wrappers.
It also provides access to the module's `Ctx`.

This allows doing this like setting some userdata in the context, and accessing it during both module declaration and evaluation.
In particular, if that userdata is a HashMap, the module name can be used as a key to access module specific userdata, so that one `ModuleDef` can be shared across different modules.
This is useful in situations where the native modules can only be defined at runtime.

### Checklist

### DIFF
--- a/core/src/value/module.rs
+++ b/core/src/value/module.rs
@@ -143,6 +143,11 @@ impl<'js> Declarations<'js> {
         unsafe { qjs::JS_AddModuleExport(self.0.ctx.as_ptr(), self.0.as_ptr(), name.as_ptr()) };
         Ok(self)
     }
+
+    /// Returns the module being declared.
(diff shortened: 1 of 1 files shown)


jprendes:
@richarddd PTAL :-)

jprendes:
The CI failure is unrelated to the changes in this PR.
Is this a known sporadic? Would restarting the failed job do the trick?

richarddd:
> The CI failure is unrelated to the changes in this PR. Is this a known sporadic? Would restarting the failed job do the trick?

Yupp. I got sick of it and found the root cause. It's qemu thread emulation that occasionally bugs. Made a PR for it

PR #631, opened 2026-02-22
fix: spec compliant iterator

richarddd:
### Description of changes

Iterators produced by Iterable now have the correct prototype chain through IteratorPrototype, with next as enumerable/writable/configurable and `[Symbol.iterator]()` returning `this`. Added `Iterable::from_fn` for closure-based  iterators that can reference live data, which is needed for Web API methods like Headers.keys(). Removed the old two-level iterable/iterator indirection. Iterable now produces an iterator object directly.

### Checklist


core/src/value/iterable.rs
```diff
@@ -1,116 +1,182 @@
...
+impl<I: IntoIterator> From<I> for Iterable<IteratorWrapper<I::IntoIter>> {
     fn from(iter: I) -> Self {
-        Iterable(iter)
+        Iterable(IteratorWrapper(Some(iter.into_iter())))
     }
 }
 
-impl<'js, I, T> IntoJs<'js> for Iterable<I>
+/// Wrapper that adapts a Rust `Iterator` into a `FnMut() -> Option<T>`.
+pub struct IteratorWrapper<I>(Option<I>);
```
Sytten: I have an alternative proposition:
```rust
trait IterableFn<T> {
   fn call(&mut self) -> Option<T>
}

impl<T, F> IterableFn<T> for F
   where F: FnMut() -> Option<T> 
{
   fn call(&mut self) -> Option<T> {
     self()
   }
}

impl<T, I> IterableFn<T> for I
   where I: Iterator<Item = T> 
{
   fn call(&mut self) -> Option<T> {
     self.next()
   }
}
```
Then you can have a single IntoJs implementation and it allows external users to define other implementations.

PR #648, opened 2026-03-14
feat(sys): allow skipping WASI SDK setup

diogotcorreia:
### Description of changes

Currently rquickjs-sys forces the WASI SDK to be used. However, in [nixpkgs](https://github.com/NixOS/nixpkgs) we do not use the WASI SDK but instead use the WASI libc directly. During our builds, `CC` and `AR` are already pointing to the correct binaries.
At the moment we are creating a temporary structure that resembles the WASI SDK just so the build succeeds, but it would be better if we didn't have to.

Therefore I added an env variable (`RQUICKJS_SYS_NO_WASI_SDK`, naming suggestions are welcome) that skips setup of the WASI SDK and places the burden of configuring the environment on whoever is building instead.

### Checklist

- [ ] ~~Created unit tests for my feature if needed~~


diogotcorreia:
I've tested this change by building https://github.com/extism/js-pdk with Nix and it works fine.

Issue #285, opened 2024-03-28
Macro to define data objects


Sytten:
I had this idea since I often use "data objects" to get multiple inputs in a JS function.

Given:
```rust
#[derive(Clone, Trace)]
#[rquickjs::data]
pub struct FindingSpec {
    title: String,
}
```

This would automatically implements
```rust
impl<'js> FromJs<'js> for FindingSpec {
    fn from_js(ctx: &Ctx<'js>, value: Value<'js>) -> Result<Self> {
        let obj = value.get::<Object<'js>>()?;
        Ok(Self {
            title: obj.get("title")?,
        })
    }
}
```

An alternative to also allow `new FindingSpec()` would be:
```rust
#[derive(Clone, Trace)]
#[rquickjs::class(data)]
pub struct FindingSpec {
    title: String,
}
```

Which would implement:
```rust
impl<'js> rquickjs::FromJs<'js> for FindingSpec
    where
        for<'a> rquickjs::class::impl_::CloneWrapper<'a, Self>:
            rquickjs::class::impl_::CloneTrait<Self>,
    {
        fn from_js(
            ctx: &rquickjs::Ctx<'js>,
            value: rquickjs::Value<'js>,
        ) -> rquickjs::Result<Self> {
            use rquickjs::class::impl_::CloneTrait;
            if let Ok(value) = rquickjs::class::Class::<Self>::from_js(ctx, value) {
               let borrow = value.try_borrow()?;
               return Ok(rquickjs::class::impl_::CloneWrapper(&*borrow).wrap_clone())
            }
            if let Ok(value) = value.get::<Object<'js>>() {
                return Ok(Self {
                   title: obj.get("title")?,
                })
            }
            Err(/* Unsure what goes here */)
        }
    }
```

Happy to implement it @DelSkayn if you like the idea!

Sytten:
Any preference on a pure data object vs class based?

DelSkayn:
The class one conflicts with the existing class macro. 

When you currently tag a class with the class attribute it will also implement FromJs. So a class with `#[derive(FromJs)]` and `#[class]` will cause an error for conflicting FromJs implementations.

PR #660, opened 2026-04-03
Add plain-data FromJs and IntoJs derives

ropoctl:
### Issue # (if available)
https://github.com/DelSkayn/rquickjs/issues/285

### Description of changes

Add plain-data FromJs and IntoJs derives

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `FromJs` and `IntoJs` derive macros for plain-data structs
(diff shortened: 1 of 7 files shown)


richarddd:
Thanks for the PR @ropoctl. How do we handle conflict from original issue?
> When you currently tag a class with the class attribute it will also implement FromJs. So a class with #[derive(FromJs)] and #[class] will cause an error for conflicting FromJs implementations.

https://github.com/DelSkayn/rquickjs/issues/285#issuecomment-2032086308

ropoctl:
Good catch. I don't think it actually makes sense to combine the two: #[class] means "this Rust type round-trips through a Class<Self> JS instance" and #[derive(FromJs/IntoJs)] means "this Rust type round-trips through a plain object/array." A type that wanted both would be ambiguous about which representation it lives as on the JS side.

I'll document the restriction on both the #[class] rustdoc and the new derives, and I can add a check in #[class] that produces a targeted compile error when FromJs/IntoJs appear in the struct's derives, so users don't run into a bare E0119. Sound good?

ropoctl:
Updated with those checks

PR #669, opened 2026-04-14
sys: let bindgen handle target triple mapping

amaanq:
### Description of changes

The build script manually passed `--target={rust_triple}` to clang via `.clang_args()`, bypassing bindgen's `rust_to_clang_target()` which correctly maps Rust triples to clang-compatible ones (e.g. `riscv64gc-unknown-linux-gnu` to `riscv64-unknown-linux-gnu`, and `*-windows-gnullvm` to `*-windows-gnu`). This broke builds on riscv64 because clang does not recognize the `gc` ISA extension suffix in the target triple.

The manual `--target` and the gnullvm workaround from #506 are both unnecessary since bindgen 0.65+ handles these mappings internally.

(My use case is getting riscv64 bindings generated for tree-sitter now that we have a risc-v runner in theory)

### Checklist

### DIFF
--- a/sys/build.rs
+++ b/sys/build.rs
@@ -293,7 +293,7 @@ where
 }
 
 #[cfg(feature = "bindgen")]
-fn bindgen<'a, D, H, X, K, V>(out_dir: D, header_file: H, defines: X, mut add_cflags: Vec<String>)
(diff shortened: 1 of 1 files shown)


richarddd (approved):
LGTM

PR #675, opened 2026-04-22
Add #[qjs(prop)] for data properties on class prototypes

richarddd:
Adds a `#[qjs(prop)]` method attribute that declares a JS data property (evaluated at class registration time) on the prototype, enabling spec-correct `Symbol.toStringTag` and similar Web IDL hooks without the `FromJs<Self>` receiver accessor `#[qjs(get)]` requires.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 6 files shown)


Sytten (changes requested):
Code is fine but you need to add documentation to the macro at minimum. An example would also be welcome with the different setter, getter, prop, configurable (unsure what that does), writable, etc.

Sytten (approved):

PR #677, opened 2026-04-23
Reject JsClass types as class fields; guide users to Class<'js, T>

### LINKED ISSUE #532: Nested Class changes ignored because generated getter clones
The assert in the following code fails because the auto-generated getter from `rquickjs::class` clones the field, which makes changes to nested attributes get ignored.

https://github.com/DelSkayn/rquickjs/blob/36008247d1a2ad44f64b58f222152e70275e1325/macro/src/fields.rs#L247-L253

```toml
[package]
name = "rquickjs-bug"
version = "0.1.0"
edition = "2024"

[dependencies]
rquickjs = { version = "0.9.0", features = ["macro"] }
```

```rust
use rquickjs::{class::Trace, Class, JsLifetime, Object};

#[derive(Trace, JsLifetime, Default, Clone)]
#[rquickjs::class]
struct A {
    #[qjs(get)]
    b: B,
}

#[derive(Trace, JsLifetime, Default, Clone)]
#[rquickjs::class]
struct B {
    #[qjs(get, set)]
    c: String,
}

fn main() {
    let rt = rquickjs::Runtime::new().unwrap();
    let ctx = rquickjs::Context::full(&rt).unwrap();

    ctx.with(|ctx| {
        let a = A {
            b: B {
                c: "my string".into(),
            },
        };
        let cls = Class::instance(ctx.clone(), a).unwrap();
        let val = cls.into_value();
        let obj = val.as_object().unwrap();

        // Set obj.b.c = "other string"
        obj.get::<_, Object>("b")
            .unwrap()
            .set("c", "other string")
            .unwrap();

        // Get obj.b.c
        let c = obj
            .get::<_, Object>("b")
            .unwrap()
            .get::<_, String>("c")
            .unwrap();

        // Bug: obj.b.c still has "my string"
        assert_eq!(c.as_str(), "other string");
    });
}
```

richarddd:
### Issue # (if available)

Fixes #532

### Description of changes

`#[rquickjs::class]` fields typed directly as a `JsClass` silently dropped nested mutations because the generated getter cloned the value and wrapped the clone in a fresh class instance. Now such fields produce a clear compile error (`E0277` via `#[diagnostic::on_unimplemented]`) directing users to wrap the field in `Class<'js, T>` to share the underlying cell. Implemented with autoref-specialization: a `JsClassFieldCheck<T>` helper whose inherent `check` method is bounded on an unimplementable marker trait `NotAJsClassField` when `T: JsClass`, and otherwise falls back to a trivial trait impl.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed promise future aborting on stale pending exceptions by only bailing on uncatchable errors (e.g. interrupt handler)
 - Fixed iterators to use correct IteratorPrototype chain
 - Fixed a latent ABI layout vulnerability in `JS_NewPromiseCapability` FFI boundary by replacing tuple with strictly compatible array
+- `#[rquickjs::class]` now rejects fields whose type implements `JsClass` with a clear compile error; such fields silently dropped nested mutations because the generated getter cloned the value. Wrap the field in `Class<'js, T>` instead #[532](https://github.com/DelSkayn/rquickjs/issues/532)
 
 ## [0.11.0] - 2025-12-16
 
--- a/core/src/class/impl_.rs
+++ b/core/src/class/impl_.rs
@@ -1,6 +1,6 @@
 //! Helper classes and functions for use inside the macros.
 
-use crate::{value::Constructor, Ctx, Object, Result};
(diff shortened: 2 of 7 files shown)


Sytten (approved):
Not really a fan of the implementation that adds useless code but I dont see a better way.

richarddd:
This is a bit hacky but wont affect build if you avoid this pattern.

PR #680, opened 2026-04-24
fix: make async runtime and context Send/Sync without parallel feature

### LINKED ISSUE #495: Why does async_with need the future to implement Send?

richarddd:
### Issue #

Fixes #495

### Description of changes

`AsyncRuntime`, `AsyncContext`, and `WithFuture` are now unconditionally `Send`/`Sync` so `async_with` can be used with multi-threaded executors without requiring the `parallel` feature, since the underlying `Arc<async_lock::Mutex<_>>` is already thread-safe.

### Checklist


Sytten:
I am very unconvinced about that one. Parallel also enables a bunch of things like Mutex for Ref, JS_UpdateStackTop calls, etc. that are required for it to work correctly multi-threaded.

richarddd:
Yeah true, let's ditch it and close the issue. Better to be safe here

PR #683, opened 2026-04-27
feat: add ArrayBuffer::from_raw_parts and from_raw_parts_shared for external buffers

richarddd:
### Issue #

N/A

### Description of changes

Adds `ArrayBuffer::from_raw_parts` (plain `ArrayBuffer`) and `ArrayBuffer::from_raw_parts_shared` (`SharedArrayBuffer`) for wrapping caller-owned memory with a custom `JSFreeArrayBufferDataFunc`, enabling zero-copy exposure of Rust-owned byte storage (`Arc<[u8]>`, `bytes::Bytes`, memory maps, …) to JS; `drop_fn` is invoked synchronously on construction failure so the caller's buffer is released exactly once.

### Checklist

### DIFF
--- a/core/src/value/array_buffer.rs
+++ b/core/src/value/array_buffer.rs
@@ -1,4 +1,6 @@
 use crate::{qjs, Ctx, Error, FromJs, IntoJs, JsLifetime, Object, Result, Value};
+use alloc::boxed::Box;
+use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::{
     ffi::c_void,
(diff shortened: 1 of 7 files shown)


core/src/value/array_buffer.rs
```diff
@@ -90,6 +90,66 @@ impl<'js> ArrayBuffer<'js> {
...
+    ///
+    /// `ptr` must point to `len` bytes of valid memory for the entire lifetime
+    /// of the returned buffer and any views of it, and `drop_fn` must
+    /// correctly release that memory.
+    pub unsafe fn from_raw_parts(
+        ctx: Ctx<'js>,
+        ptr: *mut u8,
+        len: usize,
+        drop_fn: qjs::JSFreeArrayBufferDataFunc,
+        opaque: *mut c_void,
```
Sytten: This is too low level of an API to be public IMO, we never expose qjs objects in the public API.
I think we should accept a higher level function like we do in `pub unsafe fn set_promise_hook(&mut self, hook: Option<PromiseHook>)`
You can keep the ptr and len though one might prefer a trait there with common implementations.

Sytten (changes requested):

core/src/value/array_buffer.rs
```diff
@@ -90,6 +151,135 @@ impl<'js> ArrayBuffer<'js> {
...
+    ///
+    /// `ptr` must point to `len` bytes of valid memory for the entire lifetime
+    /// of the returned buffer and any views of it. The `drop` closure must
+    /// release that memory correctly.
+    pub unsafe fn from_raw_parts(
+        ctx: Ctx<'js>,
+        ptr: *mut u8,
+        len: usize,
+        drop: ArrayBufferDrop,
+    ) -> Result<Self> {
```
Sytten: Did you forget to remove this API or did you intend to keep the unsafe interface? It would be new for the project.

core/src/value/array_buffer.rs
```diff
@@ -12,6 +14,65 @@ use core::{
...
+/// satisfies the `Send` requirement only when the `parallel` feature is
+/// enabled.
+#[cfg(not(feature = "parallel"))]
+pub trait DropSend {}
+#[cfg(not(feature = "parallel"))]
+impl<T> DropSend for T {}
+#[cfg(feature = "parallel")]
+pub trait DropSend: Send {}
+#[cfg(feature = "parallel")]
+impl<T: Send> DropSend for T {}
```
Sytten: We already have `ParallelSend` for that

core/src/value/array_buffer.rs
```diff
@@ -90,6 +151,135 @@ impl<'js> ArrayBuffer<'js> {
...
+        immutable: bool,
+    ) -> Result<Self> {
+        extern "C" fn shim(_rt: *mut qjs::JSRuntime, opaque: *mut c_void, _ptr: *mut c_void) {
+            unsafe {
+                let boxed: Box<ArrayBufferDrop> = Box::from_raw(opaque as *mut ArrayBufferDrop);
+                (*boxed)();
+            }
+        }
+
+        let opaque = Box::into_raw(Box::new(drop)) as *mut c_void;
```
Sytten: ArrayBufferDrop is already boxed

Sytten (changes requested):

Sytten (approved):

PR #687, opened 2026-05-12
fix: Return Result from Value::new_big_int

### LINKED ISSUE #585: Handle errors in Value new_X
For example BigInt::from_i64 will check for exceptions but not Value::new_big_int

richarddd:
Closes #585 by returning `Result<Value>` from `Value::new_big_int` so QuickJS allocation failures surface to the caller instead of silently producing a value that wraps `JS_EXCEPTION`.


Sytten (approved):
Thanks, sorry for delay

PR #689, opened 2026-05-12
Refactor CI to use cargo-zigbuild and split bindings workflow

richarddd:
Replaces the per-target apt and musl-cross toolchain installs with cargo-zigbuild, moves bindings regeneration into a dedicated workflow triggered by submodule or build-script changes, and drops the update-bindings branch trigger that caused self-triggering runs and wrong PR base branches.


.github/workflows/bindings.yml
```diff
@@ -0,0 +1,134 @@
...
+      - name: Set up Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          target: ${{ matrix.target }}
+          components: rust-src
+
+      - name: Set up zig
+        if: matrix.builder == 'zigbuild'
+        uses: mlugg/setup-zig@v2
```
Sytten: please pin rando actions

.github/workflows/bindings.yml
```diff
@@ -0,0 +1,134 @@
...
+        with:
+          version: 0.16.0
+
+      - name: Install cargo-zigbuild
+        if: matrix.builder == 'zigbuild'
+        uses: taiki-e/install-action@v2
```
Sytten: same

.github/workflows/bindings.yml
```diff
@@ -0,0 +1,134 @@
...
+        with:
+          tool: cargo-zigbuild
+
+      - name: Set up msys2
+        if: matrix.builder == 'msys2'
+        uses: msys2/setup-msys2@v2
```
Sytten: same

Sytten (changes requested):
At some point it would be easier to use nix LOL

Sytten (approved):

PR #706, opened 2026-06-29
Fix reference cycle leak in Constructor::new_prototype

### LINKED ISSUE #705: [Bug] `Constructor::new_prototype` leaks the constructor + prototype (assertion in `JS_FreeRuntime` on `Runtime` drop)
## The problem

`Constructor::new_prototype` captures the `prototype` `Object` by value inside the constructor closure. Because `JS_SetConstructor` also wires `prototype.constructor = func`, the constructor and prototype form a reference cycle. The captured handle is an owned refcount the GC's mark/sweep can't trace. The cycle survives `JS_FreeRuntime`, which trips:

```
Assertion failed: (list_empty(&rt->gc_obj_list)), function JS_FreeRuntime, file quickjs.c
```

(line 2348 on the quickjs bundled with 0.12.0).

This aborts in debug **and** release: `rquickjs-sys`'s build never defines `NDEBUG`, and neither does the `cc` crate, so the C `assert` stays compiled in.

Reproduces on **0.12.0** and at least as far back as **0.9.0**.

## Reproducible example

`Cargo.toml`:

```toml
[dependencies]
rquickjs = "0.12"
```

`src/main.rs`:

```rust
use rquickjs::function::Constructor;
use rquickjs::{Context, Ctx, Object, Runtime, Value};

fn install<'js>(ctx: &Ctx<'js>) -> rquickjs::Result<()> {
    let proto = Object::new(ctx.clone())?;
    let ctor = Constructor::new_prototype(
        ctx,
        proto,
        |ctx: Ctx<'js>| -> rquickjs::Result<Object<'js>> { Object::new(ctx.clone()) },
    )?;
    ctx.globals().set("Foo", ctor)?;
    Ok(())
}

fn main() {
    ctx.with(|ctx| {
        install(&ctx).unwrap();
        let _: Value = ctx.eval("new Foo()").unwrap();
    });
    drop(ctx);
    drop(rt); // <-- assertion fires here
    println!("clean shutdown (no leak)");
}
```

```
$ cargo run
Assertion failed: (list_empty(&rt->gc_obj_list)), function JS_FreeRuntime, file quickjs.c, line 2348.
[1]    abort      cargo run
```

(`new Foo()` isn't even required — just creating the constructor and dropping the runtime leaks; the eval is included to mirror real usage.)

### Root cause

`new_prototype` keeps an owned clone of the prototype for the "called without `new`" fallback:

```rust
pub fn new_prototype<F, P>(ctx: &Ctx<'js>, prototype: Object<'js>, f: F) -> Result<Self> {
    let proto_clone = prototype.clone();                       // <-- owned capture
    let func = Box::new(move |params| {
        let proto = this
            .as_function()
            .map(|func| func.get(PredefinedAtom::Prototype))
            .unwrap_or_else(|| Ok(Some(proto_clone.clone())))?; // fallback uses the capture
        ...
    });
    // JS_SetConstructor wires func.prototype = prototype AND prototype.constructor = func
}
```

The boxed closure is stored as the constructor function object's opaque, so:

```
func ──opaque──▶ closure ──owns──▶ prototype   (refcount++, untraced by GC)
  ▲                                    │
  └──────── prototype.constructor ◀────┘        (JS_SetConstructor)
```

Mark/sweep can collect the `func ⇄ prototype` cycle, but the extra refcount held by the Rust closure makes `prototype` look externally reachable, so the cycle is never freed.

`Constructor::new_class` does **not** have this problem: its fallback re-derives the prototype from the class registry (`.unwrap_or_else(|| Class::<C>::prototype(&ctx))`) instead of capturing an owned handle. The difference between the two fallbacks is the bug.

### Why this isn't the same as #370 / #448

Both of those hit the same assertion, but the capture is in *user* code, and the resolution is: don't capture but take `ctx` / the value as a closure parameter. That guidance can't be applied here: the capture is inside `new_prototype` itself, so a caller using the documented public API ("Useful if the function does not return a Rust class") gets an unavoidable leak/abort.

### Suggested fix

Make `new_prototype` not hold an owned strong ref to the prototype. Options:

1. Mirror `new_class`: drop the `proto_clone` fallback and, when `new.target` isn't a function (the no-`new` call), fall back to the default object prototype rather than the captured one.
2. Hold the fallback prototype somewhere the GC traces (or as a weak ref) instead of an owned `Object` in the closure.

At minimum, document that `new_prototype` leaks the constructor/prototype across `Runtime` teardown, and point users at the `new.target` pattern below.

### Workaround (no capture, `instanceof` still holds)

Build the constructor as a plain `Function` with the constructor bit and read the prototype off `new.target` instead of capturing it. The only cycle left (`ctor.prototype ⇄ prototype.constructor`) is pure JS, which the cycle collector frees:

```rust
use rquickjs::function::This;
use rquickjs::{Ctx, Function, Object, Value};

fn install<'js>(ctx: &Ctx<'js>) -> rquickjs::Result<()> {
    let proto = Object::new(ctx.clone())?;
    // ... set methods on `proto` ...
    let ctor = Function::new(
        ctx.clone(),
        |this: This<Value<'js>>, ctx: Ctx<'js>| -> rquickjs::Result<Object<'js>> {
            let inst = Object::new(ctx.clone())?;
            if let Some(f) = this.0.as_function() {
                if let Some(p) = f.get::<_, Option<Object<'js>>>("prototype")? {
                    inst.set_prototype(Some(&p))?;
                }
            }
            Ok(inst)
        },
    )?
    .with_constructor(true);
    ctor.set("prototype", proto.clone())?;
    proto.set("constructor", ctor.clone())?;
    ctx.globals().set("Foo", ctor)?;
    Ok(())
}
```

This exits cleanly and `(new Foo()) instanceof Foo === true`.

### Environment

- rquickjs 0.12.0
- macOS arm64

richarddd:
### Description of changes

`Constructor::new_prototype` captured an owned clone of the prototype inside the constructor closure (stored as the function object's opaque data); combined with the `prototype.constructor = func` wiring done by `JS_SetConstructor`, this formed an untraceable reference cycle that survived `JS_FreeRuntime` and tripped the `list_empty(&rt->gc_obj_list)` assertion (`quickjs.c` line 2348), aborting on drop of the runtime. The fix mirrors `new_class` by deriving the prototype dynamically from the called function (`params.function().prototype`) in the fallback path instead of capturing an owned handle, breaking the cycle.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed `Constructor::new_prototype` leaking a reference cycle that aborted `JS_FreeRuntime` with a `gc_obj_list` assertion failure
+
 ### Security
 
 ## [0.12.0] - 2026-05-26
--- a/core/src/value/function.rs
+++ b/core/src/value/function.rs
@@ -240,14 +240,16 @@ impl<'js> Constructor<'js> {
     where
         F: IntoJsFunc<'js, P> + 'js,
     {
-        let proto_clone = prototype.clone();
         let func = Box::new(move |params: Params<'_, 'js>| -> Result<Value<'js>> {
             params.check_params(F::param_requirements())?;
             let this = params.this();
+
             let proto = this
-                .as_function()
+                .into_function()
+                .or_else(|| params.function().into_function())
                 .map(|func| func.get(PredefinedAtom::Prototype))
-                .unwrap_or_else(|| Ok(Some(proto_clone.clone())))?;
+                .transpose()?
+                .flatten();
 
             let res = f.call(params)?;
             res.as_object()
@@ -815,6 +817,32 @@ mod test {
                 .unwrap();
             let n: u32 = ctx.eval("log(\"foo\") + log(\"bar\")").unwrap();
             assert_eq!(n, 3);
+        });
+    }
+
+    #[test]
+    fn constructor_new_prototype_no_leak() {
+        use crate::function::Constructor;
+
+        // Creating a constructor via `new_prototype` used to capture an owned clone of the
+        // prototype inside the constructor closure. Combined with the
+        // `prototype.constructor = func` wiring done by `JS_SetConstructor`, that formed a
+        // reference cycle the GC could not collect, which tripped the
+        // `list_empty(&rt->gc_obj_list)` assertion in `JS_FreeRuntime` when the runtime was
+        // dropped. This test reproduces that teardown (the runtime is dropped at the end of
+        // `test_with`) and must shut down cleanly.
+        test_with(|ctx| {
+            fn make<'js>(ctx: Ctx<'js>) -> Result<Object<'js>> {
+                Object::new(ctx)
+            }
+            let proto = Object::new(ctx.clone()).unwrap();
+            let ctor = Constructor::new_prototype(&ctx, proto, make).unwrap();
+            ctx.globals().set("Foo", ctor).unwrap();
+
+            // Mirror real usage: actually construct an instance and verify its prototype is
+            // wired up correctly via the fallback path.
+            let is_instance: bool = ctx.eval("var f = new Foo(); f instanceof Foo").unwrap();
+            assert!(is_instance);
         });
     }
 }


Sytten (approved):

Sytten:
@richarddd Can you handle the patch release

PR #714, opened 2026-07-22
Add Object::typed_array_type for full typed array class detection

richarddd:
`is_typed_array::<T>()` can only match typed array classes that have a `TypedArrayItem` impl. `Uint8ClampedArray` has none (it shares its element type `u8` with `Uint8Array`), so there is currently no safe way to detect it, or to distinguish it from a `DataView` when both are reached through a backing `ArrayBuffer`. The only workarounds are calling `JS_GetTypedArrayType` via the raw `qjs` bindings or reading a spoofable `Symbol.toStringTag`.

This adds:
- `TypedArrayType`, a public enum covering every typed array class the engine reports, including `Uint8Clamped` and `Float16`.
- `Object::typed_array_type(&self) -> Option<TypedArrayType>`, a safe wrapper over `JS_GetTypedArrayType` returning `None` for non-typed-array objects.

`is_typed_array` is unchanged. Added a unit test covering all classes plus `DataView`/plain-object negatives.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -57,7 +57,9 @@ pub use context::MultiWith;
 #[cfg(feature = "futures")]
 #[cfg_attr(feature = "doc-cfg", doc(cfg(feature = "futures")))]
 pub use runtime::AsyncRuntime;
-pub use value::{ArrayBuffer, ArrayBufferSource, Iterable, IterableFn, JsIterator, TypedArray};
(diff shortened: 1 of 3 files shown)


richarddd:
Closing: this only replaces a single, already-robust raw JS_GetTypedArrayType call downstream. Not worth a public API addition.

PR #723, opened 2026-08-07
Update quickjs-ng to 2c620e4 and fix ArrayBuffer for new JS_NewArrayBuffer signature

richarddd:
quickjs-ng added a max_len param to JS_NewArrayBuffer and changed the free callback into a realloc callback, and moved ref_count out of the object body into the allocator header.

Updated both call sites and the callback signatures, regenerated bindings, and dropped the ref_count check in drop_context since that field is no longer readable.

Updated to quickjs-ng 0.16.2 (2c620e4), which includes several bug fixes on top of 0.16.0 with no further API changes.

Supersedes #722.


core/src/context/async.rs
```diff
@@ -85,16 +85,6 @@ impl DropContext for AsyncRuntime {
...
-                    let p =
-                        unsafe { &mut *(ctx.as_ptr() as *mut crate::context::ctx::RefCountHeader) };
-                    if p.ref_count <= 1 {
-                        // Lock was poisoned, this should only happen on a panic.
-                        // We should still free the context.
-                        // TODO see if there is a way to recover from a panic which could cause the
-                        // following assertion to trigger
-                        #[cfg(feature = "std")]
-                        assert!(std::thread::panicking());
-                    }
```
Sytten: If there a way to detect lock poisoning or do panic recovery?

core/src/value/array_buffer.rs
```diff
@@ -197,11 +208,16 @@ impl<'js> ArrayBuffer<'js> {
             _rt: *mut qjs::JSRuntime,
             opaque: *mut c_void,
             _ptr: *mut c_void,
-        ) {
+            size: qjs::size_t,
+        ) -> *mut c_void {
+            if size != 0 {
+                return core::ptr::null_mut();
+            }
```
Sytten: Add a comment to indicate that we only need to support the size 0 case (free command) and that other sizes are not expected since we set max_len to 0 (fixed) in the New function.

core/src/value/array_buffer.rs
```diff
@@ -94,19 +94,30 @@ impl<'js> ArrayBuffer<'js> {
...
         let size = src.len() * size_of::<T>();
 
-        extern "C" fn drop_raw<T>(_rt: *mut qjs::JSRuntime, opaque: *mut c_void, ptr: *mut c_void) {
+        extern "C" fn drop_raw<T>(
+            _rt: *mut qjs::JSRuntime,
+            opaque: *mut c_void,
+            ptr: *mut c_void,
+            size: qjs::size_t,
+        ) -> *mut c_void {
+            if size != 0 {
```
Sytten: Same comment

core/src/value/array_buffer.rs
```diff
@@ -211,12 +227,19 @@ impl<'js> ArrayBuffer<'js> {
...
                 opaque,
                 is_shared,
             );
             if let Err(e) = ctx.handle_exception(val) {
-                shim::<F>(qjs::JS_GetRuntime(ctx.as_ptr()), opaque, ptr as *mut c_void);
+                shim::<F>(
+                    qjs::JS_GetRuntime(ctx.as_ptr()),
+                    opaque,
+                    ptr as *mut c_void,
+                    0,
```
Sytten: Add a constant like `const FREE: qjs::size_t = 0;`

core/src/value/array_buffer.rs
```diff
@@ -211,12 +227,19 @@ impl<'js> ArrayBuffer<'js> {
                 ctx.as_ptr(),
                 ptr,
                 len as _,
+                // fixed-length buffer, not resizable
+                0,
```
Sytten: Add a constant like `const FIXED_SIZE: qjs::size_t = 0;`

core/src/context/async.rs
Sytten: Did we abandon the idea of supporting bellard quickjs?

core/src/value/array_buffer.rs
```diff
@@ -637,6 +660,69 @@ mod test {
         });
     }
 
+    #[test]
+    fn transfer_to_different_length_preserves_vec_buffer() {
```
Sytten: In theory we could support the transfer by allocating a new vec if we want.

Sytten (changes requested):
Couple of comments

core/src/context/async.rs
richarddd: https://github.com/richarddd/rquickjs/pull/3/changes

I dont know it seems quite risky as the ABI has diverted quite a bit, not to bad tho.

core/src/value/array_buffer.rs
```diff
@@ -89,26 +94,63 @@ impl<'js> ArrayBuffer<'js> {
...
+                return core::ptr::null_mut();
+            }
+            let new_len = (size as usize).div_ceil(elem_size);
+
+            if new_len > capacity {
+                if vec.try_reserve_exact(new_len - capacity).is_err() {
+                    return core::ptr::null_mut();
+                }
+            } else if new_len < capacity {
+                unsafe { vec.set_len(new_len) };
```
Sytten: Technically this is not safe because the current method allows:
```rust
#[derive(Copy, Clone)]
struct Point {
   x: i32,
   y: i32,
}

impl Drop on Point {
  fn drop(&mut self) {
     println!("Hello");
  }
}

let vec = vec![Point { x:1, y:2 }];
ArrayBuffer::new(ctx, vec);
```

If you use sel_len instead of truncate then the Drop is not called.

Sytten (changes requested):

core/src/value/array_buffer.rs
richarddd: nice catch!

richarddd: Fixed, switched to vec.truncate(new_len).

Sytten (approved):

PR #724, opened 2026-08-07
Fix deadlock when dropping a Context while the runtime lock is held

richarddd:
Mut::try_lock calls the blocking lock under the parallel feature, so dropping a Context while the same thread already holds the runtime lock deadlocks.

Track the locks held per thread so drop_context can free directly when it is already on that thread, and leave genuine cross-thread contention blocking as before.

Reachable without any explicit nested drop: a JsClass holding a Context deadlocks when its finalizer runs inside `with`. Works without the parallel feature, hangs with it.

### DIFF
--- a/core/src/context/base.rs
+++ b/core/src/context/base.rs
@@ -1,5 +1,4 @@
 use super::{
-    ctx::RefCountHeader,
     intrinsic,
     owner::{ContextOwner, DropContext},
     ContextBuilder, Intrinsic,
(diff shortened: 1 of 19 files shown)


richarddd:
@DelSkayn would like your eyes on this one if possible

core/src/runtime/raw.rs
```diff
@@ -115,6 +119,44 @@ pub(crate) struct RawRuntime {
...
+        f.write_str("PendingFree")
+    }
+}
+
+#[cfg(feature = "parallel")]
+impl PendingFree {
+    pub(crate) fn push(&self, ctx: NonNull<qjs::JSContext>) {
+        if let Ok(mut queued) = self.0.lock() {
+            queued.push(CtxPtr(ctx));
+        }
```
Sytten: No a fan of hiding a poison lock

core/src/context/base.rs
```diff
@@ -13,21 +12,27 @@ impl DropContext for Runtime {
...
-                if p.ref_count <= 1 {
-                    // Lock was poisoned, this should only happen on a panic.
-                    // We should still free the context.
-                    // TODO see if there is a way to recover from a panic which could cause the
-                    // following assertion to trigger
-                    #[cfg(feature = "std")]
-                    assert!(std::thread::panicking());
+                #[cfg(feature = "parallel")]
+                {
+                    self.pending_free.push(ctx);
```
Sytten: In the async runtime we use a `drop_send`, I just want to check if it is possible/worth it to uniformise the methods.

Sytten (approved):
To the best of my ability this is good, but see my comments please

core/src/runtime/raw.rs
richarddd: Swapped it for an mpsc channel, so there's no lock to poison.

core/src/context/base.rs
richarddd: Yep, worth it: RawRuntime now owns the channel and both Runtime and AsyncRuntime drain it through the same drain_pending_free, so the parking logic isn't duplicated anymore.

PR #731, opened 2026-08-26
Update native module examples to the current module macro

### LINKED ISSUE #483: Examples use `rquickjs::bind` but that's not available since 0.4.0
Hi,

I have been trying to make some native modules and the `rquickjs::bind` macro is not available since 0.4.0.

What's the way to do that right now? Is there any other macro I'm missing?

ChrisJr404:
The `using_macro` example files still used the `rquickjs::bind` macro that was removed back in 0.4.0, so they were left commented out of the build and only the hand written path was ever exercised. This ports them to the current `#[rquickjs::module]` macro and re-enables the macro feature toggle so both examples run through the macro path again, and I confirmed `module-loader` prints the expected output with and without the macro feature. Closes #483.

### DIFF
--- a/examples/module-loader/src/bundle.rs
+++ b/examples/module-loader/src/bundle.rs
@@ -1,14 +1,14 @@
-//#[cfg(not(feature = "macro"))]
+#[cfg(not(feature = "macro"))]
 mod hand_written;
 
-//#[cfg(feature = "macro")]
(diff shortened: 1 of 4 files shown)


richarddd (approved):

Sytten:
@richarddd Can you pin the clippy to the MSRV

Issue #736, opened 2026-09-03
TypedArray::as_bytes returns a &[u8] that is actually mutable


goffrie:
Consider this program:

```rust
use rquickjs::{Context, Runtime, TypedArray};
fn main() {
    let rt = Runtime::new().unwrap();
    let ctx = Context::full(&rt).unwrap();
    ctx.with(|ctx| {
        let buffer = TypedArray::<u8>::new(ctx.clone(), vec![1, 2, 3]).unwrap();

        let bytes: &[u8] = buffer.as_bytes().unwrap();
        let before = bytes.to_vec();

        buffer.set(0, 39).unwrap();

        let after = bytes.to_vec();
        assert_eq!(before, after);
    });
}
```

The assertion fails with

```
assertion `left == right` failed
  left: [1, 2, 3]
 right: [39, 2, 3]
```

This is highly questionable since `bytes` is supposed to be an immutable `&[u8]`.

It seems that the `as_bytes()` API is problematic since it returns a slice whose lifetime is pretty broad, tied to a borrow of the `TypedArray` handle, but nothing is stopping JS operations (like array writes) from happening during that time. I believe the same applies to the `AsRef<[T]>` impl, and both also for `ArrayBuffer`.

goffrie:
fixed by #737

PR #737, opened 2026-09-03
Mark TypedArray/ArrayBuffer byte-slice accessors unsafe

richarddd:
### Issue # (if available)

https://github.com/DelSkayn/rquickjs/issues/736

### Description of changes

Marks `TypedArray`/`ArrayBuffer` byte-slice accessors `unsafe` and removes the unsound safe `AsRef<[T]>` impls, since JS can mutate, detach, or reallocate the backing store while the returned slice is still borrowed.

### Checklist


Sytten (approved):

tustvold:
As these are now unsafe is it worth providing safe APIs for converting that copy into `Vec<T>` or similar? As far as I can tell there is no safe way to access the data anymore.

For context this is the code I am trying to write

```
// Safety: the slice is copied immediately with no JS execution in
// between, so the buffer can't be detached/reallocated while it's held.
let bytes = unsafe { ta.as_bytes() }.ok_or_else(|| Error::new_from_js("Uint8Array", "bytes"))?;
return Ok(Self::Bytes(Bytes::copy_from_slice(bytes)));
```

PR #751, opened 2026-09-14
Fix type checking for &Ctx in ExoticMethod

tralwdwd:
### Description of changes
Fixes an issue where accepting a `Ctx` parameter in an exotic method caused an error because the macro checked for `Ctx` but passed `&Ctx` when the check succeeded.

### Checklist


richarddd (approved):
Thanks!

PR #754, opened 2026-09-17
Release 0.14.0

richarddd:
Minor release: removes the unused `StaticJsFunction` trait (breaking), adds `Exception::throw_dom`/`intrinsic::DOMException` support and `wasm32-unknown-unknown` platform support, and fixes `Params::is_constructor`, `ExoticMethod`'s `&Ctx` type checking, a pending-exception leak in `Ctx::execute_pending_job`, and one in `Object::instance_of`.

See CHANGELOG.md for details.

PR #756, opened 2026-09-17
Remove unused dependencies

richarddd:
Removes deps that show up as \`unused_dependencies\`/\`unused_workspace_dependencies\` manifest warnings across CI jobs, verified against source rather than just the warning text:

- \`rquickjs\` in root \`Cargo.toml\`'s \`[workspace.dependencies]\` — nothing inherits it via \`workspace = true\` (the root package can't meaningfully depend on itself).
- \`fnv\` and \`ident_case\` in \`macro/Cargo.toml\` — no references anywhere in \`macro/src/\`.

(\`rquickjs-sys\` also shows as "unused" in some CI jobs, but that's a cargo lint blind spot, not a real issue — it's used by \`core/Cargo.toml\` via \`workspace = true\`, just not visible when cargo is invoked scoped to \`sys/Cargo.toml\` alone. Left untouched.)

Verified locally: \`cargo check --workspace\`, \`cargo clippy --all --all-targets --features full-async,bindgen\`, and the full test suite (including the macro trybuild compile tests) all pass with zero unused-dependency warnings.


richarddd:
Closing as redundant — the same fix (removing the unused `rquickjs` workspace dependency and unused `fnv`/`ident_case` deps from `macro/Cargo.toml`) was pushed directly onto `release/0.14.0` in #754.

PR #760, opened 2026-09-21
Allow internal null bytes in JS source

goffrie:
### Description of changes

JS is allowed to have `\0` bytes in it, e.g. in string literals, but `CString::new(src)` improperly raises an error. QuickJS only requires

```
/* 'input' must be zero terminated i.e. input[input_len] = '\0'. */
```

so do just that.

### Checklist


richarddd (approved):

goffrie:
Whoops, forgot that c_char != u8 on many platforms. Fixed now

Sytten:
Wont this fail then on those platform @goffrie?
We are using utf-8 bytes (meaning between 1 and 4 bytes per codepoint) and casting that into a c_char of u32 that would break things in my book.
Maybe we need a CStr::from_bytes_with_nul_unchecked or something

goffrie:
`c_char` is always either `u8` or `i8`, Rust doesn't support any platforms where that isn't the case. And `CStr::from_bytes_with_nul_unchecked` just casts `*const [u8]` to a `*const [c_char]` so it's equivalent.

